PR #47, opened 2017-05-12
Fix deepCopy() functions for IGeometry objects

### LINKED ISSUE #46: Deep copy functions for IGeometry objects don't actually deep copy
The current deepCopy() implementations for IGeometry objects do not actually perform a deep copy; instead, they just increase the reference count for each of the underlying arrays.

sdonegan:
The fix to ensure that the deepCopy() actually occurs is to make sure to call deepCopy() on the underlying arrays.  The dynamic connectivity arrays do not have a deep copy implementation yet, so the new IGeometry object may need to recomputed these. 

updates #46 
closes #46


imikejackson:
Pending the comments left in issue #48 I would rather accept a patch that includes the "forceNoAllocate" argument.

sdonegan:
Assuming my response to #48 makes sense, I can make the changes to include a forceNoAllocate.

imikejackson:
Fixed by another push.

PR #108, opened 2017-07-31
FilterPipeline now returns the DataContainerArray after execution

mmarineBlueQuartz:
Running a pipeline now returns the DataContainerArray as a std::shared_ptr as it was at the end of execution.  Doing so allows the results to be returned in a way that DREAM3D plugins can more easily use without the need to save to and read from a file every time a pipeline is run.

### DIFF
--- a/Source/SIMPLib/Common/FilterPipeline.cpp
+++ b/Source/SIMPLib/Common/FilterPipeline.cpp
@@ -290,10 +290,12 @@ void FilterPipeline::cancelPipeline()
 // -----------------------------------------------------------------------------
 //
 // -----------------------------------------------------------------------------
-void FilterPipeline::run()
+DataContainerArray::Pointer FilterPipeline::run()
 {
-  execute();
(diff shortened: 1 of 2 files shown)


Source/SIMPLib/Common/FilterPipeline.cpp
```diff
@@ -589,6 +591,8 @@ void FilterPipeline::execute()
 
   PipelineMessage completMessage("", "Pipeline Complete", 0, PipelineMessage::MessageType::StatusMessage, -1);
```
imikejackson: I think you want "completeMessage" as your variable name?

imikejackson:
Does this disrupt DREAM.3D in anyway?

mmarineBlueQuartz:
execute() and run() now return a DataContainerArray::Pointer instead of void.  Other than fixing a variable name, that's all I did.  It's compiled for me for days, and I've used DREAM3D to make pipelines without removing these changes.

PR #134, opened 2017-10-12
Fixed Enable Filter action and Disable button inconsistency

mmarineBlueQuartz:
SVPipelineFilterWidget updates the disable button when its state is changed to Disable or Ready.  Clicking the disable button also causes a preflight in order to update the DataContainerArray passed into other filters.


imikejackson:
I also had this change somewhere but we will pull yours in and fix the conflicts on the next merge/pull request.

PR #139, opened 2017-10-17
Feature/190 read hdf5 file simple

joeykleingers:
Adds ReadHDF5Dataset filter, which reads an HDF5 dataset into DREAM.3D.


Source/SVWidgetsLib/FilterParameterWidgets/ImportHDF5DatasetWidget.cpp
```diff
@@ -0,0 +1,675 @@
...
+
+  if(m_Filter != nullptr)
+  {
+    QString dsetPath = m_Filter->getDatasetPath();
+    dsetPathLE->setText(dsetPath);
+
+    QString hdf5FilePath = m_Filter->getHDF5FilePath();
+    if(hdf5FilePath.isEmpty() == false)
+    {
+      value->setText(hdf5FilePath);
```
imikejackson, line 173: What happens if the file does not exist? There needs to be a check to ensure there is a file at the location in the hdf5FilePath variable.

Just drop the filter into a pipeline. it will crash.

Source/SVWidgetsLib/FilterParameterWidgets/ImportHDF5DatasetWidget.cpp
```diff
@@ -0,0 +1,675 @@
...
+      value->show();
+      initWithFile(hdf5FilePath);
+    }
+
```
imikejackson: The code does not return if there was an issue loading up the HDF5 file it will keep going.

Source/SVWidgetsLib/FilterParameterWidgets/ImportHDF5DatasetWidget.cpp
```diff
@@ -0,0 +1,675 @@
...
+    // Automatically expand HDF groups in the viewer
+    QStringList dsetTokens = dsetPath.split("/", QString::SkipEmptyParts);
+    ImportHDF5TreeModel* treeModel = dynamic_cast<ImportHDF5TreeModel*>(hdfTreeView->model());
+    QModelIndex parentIdx = treeModel->index(0, 0);
```
imikejackson: And dies here because the model was not set to ImportHDF5TreeModel treeModel will be a nullptr.

imikejackson:
The license block needs to be updated at the head of each of the files:

/* ============================================================================
 * Copyright (c) 2017 BlueQuartz Softwae, LLC
 * All rights reserved.
 *
 * Redistribution and use in source and binary forms, with or without modification,
 * are permitted provided that the following conditions are met:
 *
 * Redistributions of source code must retain the above copyright notice, this
 * list of conditions and the following disclaimer.
 *
 * Redistributions in binary form must reproduce the above copyright notice, this
 * list of conditions and the following disclaimer in the documentation and/or
 * other materials provided with the distribution.
 *
 * Neither the names of any of the BlueQuartz Software contributors
 * may be used to endorse or promote products derived from this software without
 * specific prior written permission.
 *
 * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
 * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
 * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
 * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
 * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
 * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
 * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
 * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 * USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 *
 *
 * ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ */

imikejackson:
Also be sure the include guards match the name of the header file as best as possible. 

ImportHDF5Dataset has an include guard of _readhdf5file_h_. This will prevent collisions in the future.

PR #151, opened 2017-11-24
Create a Pipeline Summary filter.

joeykleingers:


imikejackson:
The top level license comment should be changed to remove the contract numbers and substitute the most current contract number: FA8650-15-D-5231

PR #169, opened 2018-01-25
Resolved Windows CMake error when configuring without DREAM.3D

### LINKED ISSUE #168: CMake variable does not pass Windows install path requirements
DREAM3D_PACKAGE_DEST_PREFIX is empty when running CMake over SIMPL without using DREAM.3D.  This causes a CMake error on line 347 of CMakeLists.txt for Windows because the DESTINATION is set to an invalid value.

mmarineBlueQuartz:
Changed the default value for DREAM3D_PACKAGE_DEST_PREFIX from "" to "." in CMakeLists.txt if the variable is empty or does not exist.  This is important because line 347 attempts to set that variable as the destination install path on Windows.  This leads to a CMake error.  Setting the default value to "." like in DREAM.3D's CMakeLists.txt resolves this error and allows the project to compile.

Fixes #168


imikejackson:
Did you verify that running the 'Install' target installs everything correctly?

PR #226, opened 2018-07-20
Feature/import multiple hdf5 datasets

imikejackson:


imikejackson:
We should be able to guess the component dimensions based on the dims from the file and the Geometry Dims.

imikejackson:
The widget needs to be styled a bit without messing up the Dock tabs so we may need a subclass + CSS styling.

PR #244, opened 2018-09-25
Fixes assert in setTuple in DataArray.hpp

JDuffeyBQ:


imikejackson:
Did you verify this fix is correct with a unit test? Was this code ever tested using a unit test?

JDuffeyBQ:
There wasn't a test in DataArrayTest for setTuple, but I've added one now. Also the change is the same as in commit cc32cdfd96cae725635db2484a41da9ae38c94d4, but that change was only applied to one version of the overloaded function. Without it, the last tuple can't be set without the assert firing.

Issue #253, opened 2018-10-12
Import ASCII Data Import Wizard resets data types when "Start Import at Row" is changed


imikejackson:
Create the "Import ASCII Data" filter. Select a text file to import. Select the delimiter on the first page, then on the second page of the wizard start changing the import data types from the default of "double" to something else. NOW change the "Start Import At Row" value. The import types reset to double. This probably should not happen.

PR #254, opened 2018-10-12
Import ASCII Data Import Wizard resets data types when "Start Import at Row" is changed. #253

joeykleingers:
Fixing bug where changing "Start Import at Row" resets the data types in the Import ASCII Data wizard.


imikejackson:
@joeykleingers what is the difference between clear and clearContents?

joeykleingers:
Clear clears everything, including the data types for each column, row and column headers, and column errors.  ClearContents just clears the contents of the table.  I modeled it after QTableWidget, which also has clear() and clearContents() functions.

joeykleingers:
Since changing the "Start Import at Row" value only changes which content is being shown, we only need to clear the contents instead of everything.

PR #257, opened 2018-10-30
Create more verbose error message if parsing Json Pipeline fails.

imikejackson:


mmarineBlueQuartz (changes requested):
createErrorMessages should have a default in switch(parseError.error).  You might have hit all the current values, but if more are added in later versions (or, more likely, someone assigns an invalid value to this public variable), this will not catch them.  In general, switch statements should contain a default for these reasons.  Being a public variable, parseError.error would be easy to accidentally assign a different value, which is even less secure as it is not a C++11 enum.  parseError.error could contain any integer value.

QJsonParseError parseError in createErrorMessage should be a const QJsonParseError& or const QJsonParseError unless it needs to be changed within the method.

The formatting within said switch statement is also irregular.  Either have new lines before every break, or do not have them at all.  Not having them would probably be preferable.

imikejackson:
Thanks for the feedback. All corrections have been pushed.

PR #259, opened 2018-10-30
Feature/hdf5_10

mmarineBlueQuartz:
* Fixed inconsistent uses of herr_t and hid_t between method declaration and definitions.
* Methods that would return either an herr_t or an hid_t depending on the presence of errors now only return the herr_t.
* Fixed two cases where int32_t was used instead of hid_t.  hid_t no longer maps to int32_t in HDF5 1.10
* Upgrades SIMPL to handle HDF5 1.10.3 while maintaining 1.8.20 compatibility.

### DIFF
--- a/Source/H5Support/CMakeLists.txt
+++ b/Source/H5Support/CMakeLists.txt
@@ -170,7 +170,7 @@ TARGET_INCLUDE_DIRECTORIES( ${PROJECT_NAME} PUBLIC
 
 if(BUILD_SHARED_LIBS)
   if(WIN32)
-    target_compile_definitions(${PROJECT_NAME} PUBLIC "-DH5Support_BUILT_AS_DYNAMIC_LIB -DH5_USE_18_API")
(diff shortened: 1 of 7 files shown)


imikejackson:
There should be additional arguments to the file create call to HDF5 to tell HDF5 to create a backwards compatible file. 

https://www.hdfgroup.org/2018/04/why-should-i-care-about-the-hdf5-1-10-2-release/

https://www.hdfgroup.org/wp-content/uploads/2018/04/RFC-Setting-Bounds-for-Object-Creation-in-HDF5-1.10.0-Update.pdf

Please review those documents and implement the changes necessary in H5Lite to ensure that we are still creating HDF5 1.8.x compatible files, at least for a while yet.

imikejackson:
This will at least allow HDF5 1.10.3 to be used as the HDF5 library that SIMPL links against so I am going to merge this PR.

PR #260, opened 2018-11-01
Added SVOverlayWidgetButton to SVWidgetsLib

mmarineBlueQuartz:
SVOverlayWidgetButton handles the visibility and placement of another widget in the window.  Like PopUpWidget, this serves as a way to temporarily show information or options.  Unlike PopUpWidget, the widget displayed is more permanent in that it will not disappear when the user clicks elsewhere on the screen and can be turned both on and off by the user.

### DIFF
--- a/Source/SVWidgetsLib/Widgets/SVOverlayWidgetButton.cpp
+++ b/Source/SVWidgetsLib/Widgets/SVOverlayWidgetButton.cpp
@@ -1,0 +1,620 @@
+/* ============================================================================
+ * Copyright (c) 2009-2018 BlueQuartz Software, LLC
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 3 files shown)


imikejackson:
Could you post either an image or show an example of its usage?

imikejackson:
Please adjust the License to only be copyright this year, since it did not exist prior to now.

Also remove all the DoD contact numbers except for FA8650-15-D-5231

imikejackson (commented):
Please see comments in the discussion.

mmarineBlueQuartz:
treeview-visualization
Here it is in an earlier prototype that's been placed on hold.  It's especially good for menus where shrinking the size of what is being operated on is not optimal.  A good example of this would be camera settings where the user can expand the menu to show additional options, but constantly adjusting the size of the camera preview would mess with formatting.

imikejackson:
Nice addition to the Widgets library. Merging.

PR #266, opened 2018-11-12
Enable support for using the HDF5 1.10.3 library

imikejackson:
This change will allow the creation or opening of files that are backwards
compatible with HDF5 1.8.x.


mmarineBlueQuartz (approved):
This looks good.

PR #268, opened 2018-11-17
Enable per object parallel compilation

dzenanz:
This parallelizes compilation within this library. The motivation was that 7 of my cores were sitting idle for a long time while SIMPLib was being compiled. It almost took me less time to write the patch then it did for my computer to compile it in single-threaded mode.

### DIFF
--- a/Source/SIMPLib/CMakeLists.txt
+++ b/Source/SIMPLib/CMakeLists.txt
@@ -171,6 +171,11 @@ endif()
 add_library(${PROJECT_NAME} ${LIB_TYPE} ${Project_SRCS} )
 CMP_AddDefinitions(TARGET ${PROJECT_NAME})
 
+if(MSVC)
+  # enable per object parallel compilation in this large library
+  target_compile_options(${PROJECT_NAME} PRIVATE "/MP")
+endif()
+
 #-- Add Target specific definitions
 if(WIN32 AND BUILD_SHARED_LIBS)
       target_compile_definitions(${PROJECT_NAME} PUBLIC "-DSIMPLib_BUILT_AS_DYNAMIC_LIB")


imikejackson:
I have updated the CMP project in all the correct locations to enable the functionality for all of the various libraries, plugins and executables.

dzenanz:
I looked at `CMP_AddDefinitions` macro, and decided no to put it there. In https://github.com/BlueQuartzSoftware/CMP/commit/f90fffb1be4ae0b74cab1c6e6e132929056db71d you didn't touch that macro. I was afraid that would enable object-level parallelism for all targets, both big and small. In the past, that had the effect of introducing much [higher parallelism](https://blog.kitware.com/cmake-building-with-all-your-cores/) than optimal. I don't know whether Microsoft has fixed it meanwhile. If you do a clean/rebuild of DREAM3D, how many instances of cl.exe are there? And how many cores do you have?

imikejackson:
Hmm. A lot more than 20 (I have a 10C/20T Xeon). I am not sure how VS handles this kind of Parallelism. In the past we use CMake -GUI to add the /MP directly to the CMAKE_CXX_ARGS but I have never actually looked at how many instances of cl.exe were executing at the same time.

image

imikejackson:
There were about 20 instances of msbuild.exe executing at that instance. So maybe that is more important than the number of cl.exe?

dzenanz:
Multiple MSBuild.exe instances mean there is target-level parallelism. If the targets (libraries) are of approximately the same size, target level parallelism is good enough. Object level parallelism is good when some targets are small while others are huge, or if dependencies between targets prevent parallel compilation. However, when `/MP` is enabled each library is compiled in a parallel manner, in addition to parallelism between targets. This caused (and apparently continues to cause) parallelism much higher than the number of cores. As [Ninja](https://ninja-build.org/) shows, object level parallelism can be done properly. I don't know why Microsoft doesn't fix the excess parallelism caused by `/MP` option. But until it does, it is probably best to enable `/MP` only for large targets (those with many files to be compiled).

dzenanz:
P.S. Number of MSBuild instances is not important, because it is `cl.exe` which consumes CPU time by doing the actual compilation. MSBuild's job is just to invoke it with proper flags and parameters.

imikejackson:
Can you try the latest CMP develop and see if it adversely affects your build times? The Windows builds are slow enough as it is, I don't need them getting any slower. Also, if there is a way to just add the /MP to the CMAKE_CXX_FLAGS that is how we usually do it. We could also add some dependent CMake code to perform some logic (maybe total number of source files) goes above a certain threshold then enable the object level parallelism.

dzenanz:
Adding/changing options in [CMAKE_CXX_FLAGS](https://stackoverflow.com/questions/28732209/change-default-value-of-cmake-cxx-flags-debug-and-friends-in-cmake) and friends has the side-effect of setting those options to all the projects. So `target_compile_options` is the right way to go.

The build time with all projects having `/MP` flag will probably not be much slower than optimal use of `/MP` for larger target. And it will almost certainly be faster than no `/MP`. The problem with it is that it frequently locks up the GUI so entire computer becomes less useful or even useless. And that is true even if `cl.exe` has lower priority than normal as enforced by e.g. [Process Hacker](https://processhacker.sourceforge.io/).

imikejackson:
I can't really fix the Windows scheduler. ;-) If you go with Target Parallelism you lose some of the object file parallelism and have slower compiles but possibly a responsive GUI. If you go with object parallelism then you most likely get faster compiles but a non-responsive GUI. I really am not sure which is the right way to go. Currently a base DREAM.3D compile without any extra plugins is about 5~6 minutes. When you add in ITK Image Processing those compile times explode to almost 20 Minutes on the same system. Internally we have discussed this and the issue is the link time for ITKImageProcessing taking a long time just due to the size of the library. Most of the rest of the plugins and apps are already compiled at that point and we are just waiting on the linker.

If we go with Target Level parallelism then we should probably start combining plugins so that every plugin has less than "some number" of filters so that we can optimize the CPU use. The issue is that I have systems with 4C/8T and systems with 16C/32T and one system with 28C/56T so what do we optimize for. Meanwhile on macOS and Linux ninja just "does the right thing". QtCreator is a nice IDE and will use Ninja and CMake without any issues. That is probably the only viable alternative.
 
I do get some complaints from a few of the Windows developers that their GUI locks up every now and then when compiling.

I am open to suggestions. I was happy to find a solution where I didn't have to remember to launch CMake-GUi.exe and add the /MP to the CMAKE_CXX_FLAGS

dzenanz:
I didn't expect you to fix the Visual Studio's compile scheduler, I was merely hoping that Microsoft would fix it eventually.

Adding `/MP` to the `CMAKE_CXX_FLAGS` is great for nightly and other testing builds. On developer's machines, not so much. So that's why I was proposing to enable `/MP` for `SIMPLib` and `ITKImageProcessing`, as those are the only truly big targets (along with `ImageProcessing` which somebody said is now deprecated).

imikejackson:
I personally don't mind the /MP for everything as I just find something to do while the build goes off. I can see this not working for other developers that only have a single machine to use. Maybe we add in a global cmake variable to either enable it everywhere or just in a few spots? How does that sound?

dzenanz:
If by global variable you mean a setting exposed in CMake GUI that sounds good. And it could default to having `/MP` for everything, in order to ease parallelism in scripted builds, nightlies etc.

And when `/MP` is not everywhere, then it should still be on for `SIMPLib`. And this PR is a good way to accomplish that.

imikejackson:
That is what I was thinking. I'll have to back out my changes and see how the effects the build when the user does not explicitly set the /MP in the CMAKE_CXX_FLAGS. I have a suspicion that the builds are going to take much longer. SIMPLib is large, so is SVWidgetsLib. ITKImageProcessing definitely is. DREAM3DReview is largish. OrientationLib and OrientationAnalysis Plugin is large. The Python bridge is VERY large.

I get to the point where I feel like I am micromanaging something that the system should be doing for me. Using ninja really does not help the responsiveness of the GUI unless you specifically invoke it with the -j N argument where N is 1 or 2 less than the total number of logical processors.

dzenanz:
I am using [Process Hacker](https://processhacker.sourceforge.io/) to have `cl.exe` run with lower priority. That really helps with GUI responsiveness, even when `ninja` is running. I think that `ninja` has a few more processes than cores by default.

imikejackson:
@dzenanz Do we still need to merge this PR?

dzenanz:
Yes, we could still merge this.

imikejackson:
This functionality has been integrated into the CMP project and is triggered by the call to the CMake macro "LibraryProperties()" on line 226 of SIMLib/CMakeLists.txt file.

I am going to close this PR.

PR #271, opened 2018-11-29
Remove any colon characters that end up in a file path.

imikejackson:
When users load a pipeline file that was produced on a Windows machine the
resulting path will typically have a [drive letter]:/ which can mess with macOS
and probably Linux. This fix will remove any ':' character that is still remaining
in the file path after generating an absolute path.


joeykleingers (changes requested):
Line 105 in SIMPLDataPathValidator.cpp is removing semicolons and not colons as the PR states it should do.

imikejackson:
@joeykleingers  Requested fixes are pushed

joeykleingers (approved):

joeykleingers (approved):

PR #272, opened 2018-12-05
Address VerifyPreconditions const change in ITK 5 master

thewtex:
This requires ITK 5 Beta 2 or ITK 4.13.2, both are currently to-be-released.

CC: @dzenanz


dzenanz (approved):
Looks good! I did not check whether it compiles.

imikejackson:
@joeykleingers You should merge this into any working branches that you are using for the OSD GUI application

PR #274, opened 2018-12-10
Update DataContainer and DataContainerArray to modern C++ standards

imikejackson:

### DIFF
--- a/Source/SIMPLib/DataContainers/DataContainerArray.cpp
+++ b/Source/SIMPLib/DataContainers/DataContainerArray.cpp
@@ -468,7 +468,7 @@ void DataContainerArray::removeDataContainerFromBundles(const QString& name)
 // -----------------------------------------------------------------------------
 //
 // -----------------------------------------------------------------------------
-bool DataContainerArray::renameDataContainerBundle(const QString& oldName, const QString newName)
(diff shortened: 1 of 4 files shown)


Source/SIMPLib/DataContainers/DataContainerArray.cpp
```diff
@@ -468,7 +468,7 @@ void DataContainerArray::removeDataContainerFromBundles(const QString& name)
 // -----------------------------------------------------------------------------
 //
 // -----------------------------------------------------------------------------
-bool DataContainerArray::renameDataContainerBundle(const QString& oldName, const QString newName)
+bool DataContainerArray::renameDataContainerBundle(const QString& oldName, const QString &newName)
```
mmarineBlueQuartz: newName has inconsistent formatting

mmarineBlueQuartz (approved):
Looks good.  There was one spot where a parameter is declared `const QString &newName` instead of `const QString& newName`, but that was it.

imikejackson:
Turns out that ITKImageProcessing will not link as it is looking for the DataContainer::New(const String &) which oddly the linker says is not there. I am tweaking the code again to see if we can get a compile and passing test.

PR #276, opened 2018-12-12
Create Filter to extract a Vertex Geometry from an Image or RectGrid Geometry

imikejackson:


Source/SIMPLib/CoreFilters/ExtractVertexGeometry.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+  DataContainer::Pointer dc = getDataContainerArray()->getPrereqDataContainer<AbstractFilter>(this, m_SelectedDataContainerName);
+  if(getErrorCondition() < 0)
+  {
+    return;
+  }
+
+  if(getVertexDataContainerName().isEmpty())
+  {
+    QString ss = QObject::tr("'New Data Container Name' is empty.  Please input a value.");
```
joeykleingers: Message should be updated to use Vertex Data Container Name instead of New Data Container Name.

joeykleingers (changes requested):
Line 156 error message needs to be updated.

Source/SIMPLib/CoreFilters/ExtractVertexGeometry.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+  DataContainer::Pointer dc = getDataContainerArray()->getPrereqDataContainer<AbstractFilter>(this, m_SelectedDataContainerName);
+  if(getErrorCondition() < 0)
+  {
+    return;
+  }
+
+  if(getVertexDataContainerName().isEmpty())
+  {
+    QString ss = QObject::tr("The name of the created Vertex DataContinaer is empty. Please specify a value.");
```
joeykleingers: 'DataContainer' is misspelled.

joeykleingers (changes requested):
Misspelling on line 156.

joeykleingers (approved):

PR #280, opened 2018-12-20
Misc style cleanups focusing on better styling inside of darker themes.

imikejackson:
+ Added x.png file to the SIMPL.qrc file to use in checkboxes
+ Redesigned SeparatorWidget to have a nice thin line next to the QLabel
+ Created method in SVStyle to wrap some text in HTML tags that will define the color based on the current style sheet
+ Remove Alternating row colors in the FilterListToolboxWidget
+ Properly use the StyleSheet information with sending messages to the StandardOutput widget

### DIFF
--- a/Source/SVWidgetsLib/Widgets/SVPipelineView.cpp
+++ b/Source/SVWidgetsLib/Widgets/SVPipelineView.cpp
@@ -1,3 +1,5 @@
+#include <utility>
+
 /* ============================================================================
  * Copyright (c) 2009-2016 BlueQuartz Software, LLC
  *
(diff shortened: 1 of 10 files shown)


Source/SVWidgetsLib/Widgets/SVPipelineView.cpp
```diff
@@ -419,7 +428,7 @@ void SVPipelineView::executePipeline()
   //  m_PipelineInFlight = getCopyOfFilterPipeline();
   m_PipelineInFlight = getFilterPipeline();
 
-  emit stdOutMessage("<b>Preflight Pipeline.....</b>");
+  formatStdOutputMessage("Preflight Pipeline.....", true);
```
mmarineBlueQuartz: The intent seems to be lost here.  `emit stdOutMessage(` tells me what is being done with the string being passed in without tracking down the functionality.  Even knowing otherwise, reading `formatStdOutputMessage(` feels like it's a local variable or UI feature that will stick around for a bit.  It's a bit long, but I would strongly recommend changing the method name to `emitFormattedStdOutputMessage` or something that describes what the method is doing.

mmarineBlueQuartz (changes requested):
`formatStdOutputMessage(const QString&, bool)` should be renamed.  It doesn't need to be as long as what I suggested, but it does something quite different than what I would expect when reading the line in context.

imikejackson:
I removed the method completely and in-lined the call to SVStyle to format the text. This version is definitely more readable and the intention of the lines of code are much clearer.

I also replaced 'for' loops with range based for loops where possible.

mmarineBlueQuartz (approved):
This looks good.

Issue #192, opened 2018-05-03
Add ability to store the Units of length as a piece of meta-data on the geometry.


imikejackson:
Helps users understand their data but otherwise has not effect on anything else.

imikejackson:
We could add a class enumeration for all the various length units:

https://en.wikipedia.org/wiki/Metric_prefix

PR #282, opened 2018-12-26
Add enumeration and accessors for Length Units in the IGeometry class.

imikejackson:
updates #192


mmarineBlueQuartz (approved):
This looks good.

PR #283, opened 2018-12-26
Add method to FilePathGenerator to generate RC filenames for Montage Data sets

imikejackson:

### DIFF
--- a/Source/SIMPLib/Utilities/FilePathGenerator.cpp
+++ b/Source/SIMPLib/Utilities/FilePathGenerator.cpp
@@ -36,6 +36,7 @@
 #include "FilePathGenerator.h"
 
 #include <QtCore/QDir>
+#include <QtCore/QTextStream>
 
 // -----------------------------------------------------------------------------
 //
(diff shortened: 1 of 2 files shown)


Source/SIMPLib/Utilities/FilePathGenerator.cpp
```diff
@@ -131,3 +132,57 @@ QVector<QString> FilePathGenerator::GenerateVectorFileList(int start, int end, i
   }
   return fileList;
 }
+
+// -----------------------------------------------------------------------------
+//
+// -----------------------------------------------------------------------------
+QVector<QString> FilePathGenerator::GeneratorMontageFileList(int rowStart, int rowEnd, int colStart, int colEnd, bool& hasMissingFiles, bool rcOrdering, const QString& inputPath,
```
joeykleingers: Is this function supposed to be called **Generator**MontageFileList, or **Generate**MontageFileList?

joeykleingers (commented):
Not sure that function name is properly named.

joeykleingers (changes requested):
Not sure that function name is properly named.

Source/SIMPLib/Utilities/FilePathGenerator.cpp
imikejackson: Generate would be better. I will fix that and repost a new version

joeykleingers (approved):

PR #284, opened 2018-12-27
Update length unit enums, add length units to Geometry descriptions

imikejackson:
This is just the start for this feature addition. Filters that reveal to the user or create geometries should either set the length unit or allow the user to set the length unit from the UI


Source/SIMPLib/Geometry/GeometryHelpers.cpp
```diff
@@ -34,6 +34,205 @@
...
+}
+
+// -----------------------------------------------------------------------------
+//
+// -----------------------------------------------------------------------------
+QString LengthUnit::TypeToString(IGeometry::LengthUnit t)
+{
+  QString geomType;
+
+  if(t == IGeometry::LengthUnit::Yoctometer)
```
mmarineBlueQuartz: A switch statement would  be shorter and catch any additional cases as warnings as they are introduced in the future as long as you don't include a default case.  It wouldn't work from string to enum, but it's wonderful when converting the other way.

Also, you can remove `QString geomType;`  The last line can simply read `return QString();` to make it obvious that the value was never set anywhere.

All together, this will shrink this TypeToString length to half of what it is now.

Source/SIMPLib/Geometry/GeometryHelpers.h
```diff
@@ -54,6 +54,22 @@
 namespace GeometryHelpers
 {
 
+/**
+ * @brief The LengthUnit class
+ */
+class LengthUnit
```
mmarineBlueQuartz: Do we want to make this a namespace instead?  No object of this type can exist as there is no access to the constructor, and there are only static methods available.

mmarineBlueQuartz (changes requested):
Change LengthUnit::TypeToString to use a switch statement for readability and future-proofing.

Source/SIMPLib/Geometry/GeometryHelpers.cpp
imikejackson: Made the changes along with a slightly different design that works better.

Source/SIMPLib/Geometry/GeometryHelpers.h
imikejackson: Made the changes along with a slightly different design that works better.

imikejackson:
I moved those methods inside of IGeometry which is where they belong anyways. I also moved the methods to convert between a string and the type of geometry inside of the IGeometry class to be consistent. if-thens were changed to switch where appropriate

mmarineBlueQuartz (changes requested):
Hate to be a stickler, but the IGeometry::Type and IGeometry::LengthUnit string conversion methods went from static to non-static without using any local variables.  We should probably throw static back on as the current implementation should not require an object to exist or the calls to be valid.

imikejackson:
Done

mmarineBlueQuartz (approved):
This looks good.  Capitalizing the static methods would be better for consistency, but it is not a big deal.

imikejackson:
Done. Merging.

PR #287, opened 2018-12-28
Fix runtime crashes inside of FilterPipeline and DataArrayPath

imikejackson:
Updated DataArrayPath with clang-tidy suggestions
Reverted code from previous commit in FilterPipeline that causes
a crash during runtime when using std::copy
Removed an include from AddFilterCommand and fixed up another one.


imikejackson:
All unit tests pass on Linux (Ubuntu 14.04 with GCC 4.8.5)

mmarineBlueQuartz (approved):
This looks good.

imikejackson:
UnitTests pass on Windows (MSVC 2017) and macOS (Xcode 9.x)

PR #288, opened 2018-12-28
Add REST server functionality.

joeykleingers:
* Added ability to send a pipeline with input files from client to server, run the pipeline, and send the output files from server to client.
* Added unit test to test this functionality.


imikejackson (changes requested):
Overall not too bad. I made some changes and pushed those to the branch. Implemented an internal ServerSettings class to hold the values instead of using QSettings all over the place which has side effects. Removed the Small IN100 data files and updated tests to generate proper input/output files or reference those files from the source tree instead of copying them to the binary tree.
There probably are more areas where we could use clang-tidy suggestions to update codes specifically in the case where we can pass by const reference instead of copies.

Each of the header files really should have Developer Docs added for each method to make the code more complete and understandable for the next developer.

Source/SIMPLib/Filtering/FilterPipeline.cpp
```diff
@@ -158,7 +158,17 @@ QJsonObject FilterPipeline::toJson()
 // -----------------------------------------------------------------------------
 FilterPipeline::Pointer FilterPipeline::FromJson(const QJsonObject& json, IObserver* obs)
 {
+  if (!json.contains(SIMPL::Settings::PipelineBuilderGroup) || !json[SIMPL::Settings::PipelineBuilderGroup].isObject())
+  {
+    return FilterPipeline::NullPointer();
+  }
+
   QJsonObject builderObj = json[SIMPL::Settings::PipelineBuilderGroup].toObject();
+  if (!builderObj.contains(SIMPL::Settings::NumFilters) || !builderObj[SIMPL::Settings::NumFilters].isDouble())
```
mmarineBlueQuartz, line 167: Minor detail, but you might want to run Clang Format.

Source/SIMPLib/Plugin/SIMPLPluginConstants.h
```diff
@@ -0,0 +1,81 @@
...
+#pragma once
+
+#include <QtCore/QString>
+
+namespace SIMPL
+{
+namespace JSON 
+{
+
+static const QString ErrorCode("ErrorCode");
```
mmarineBlueQuartz: It will take more effort to change this than it's worth, but consider using `extern const` instead of `static const` for namespace constants in the future.  Inside of a namespace, static will create a new SIMPL::JSON::* for each .cpp file this header is included in.  `extern` will require a .cpp file to define the variable, but you will only have one copy of each constant regardless of the number of places this header is included (including recursively through other header files that include it), which seems to be the intent here.

No need to change it now, but keep it in mind for later.

Source/SIMPLib/REST/PipelineListener.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+// -----------------------------------------------------------------------------
+//
+// -----------------------------------------------------------------------------
+std::vector<PipelineMessage> PipelineListener::getErrorMessages()
+{
+  std::vector<PipelineMessage> errorMessages;
+
+  size_t numMessages = m_Messages.size();
+  for(size_t i = 0; i < numMessages; i++)
```
mmarineBlueQuartz, line 39: Consider using `for(const PipelineMessage& message : m_Messages)` instead.  You don't have to worry about numMessages, constantly dereferencing m_Messages[i], and it's cleaner.

Source/SIMPLib/REST/PipelineListener.cpp
```diff
@@ -0,0 +1,312 @@
...
+    }
+  }
+
+  return statusMessages;
+}
+
+// -----------------------------------------------------------------------------
+//
+// -----------------------------------------------------------------------------
+std::vector<PipelineMessage> PipelineListener::getStandardOutputMessages()
```
mmarineBlueQuartz, line 91: None of these getters are changing the object in question, so they should be marked as const.  It lets the compiler optimize better as well as declaring intent.  In addition, the range-based for loop comment from above applies to all of these.

Source/SIMPLib/REST/V1Controllers/ExecutePipelineController.h
```diff
@@ -0,0 +1,91 @@
+/* ============================================================================
+ * Copyright (c) 2017 BlueQuartz Softwae, LLC
```
mmarineBlueQuartz: Randomly noticed that Software is misspelled.  Try a mass search and replace as this does not seem to be the only one.  You might as well update the years at the same time.  These files were hit at least in 2018, and fixing the company name in the copyright puts it in 2019.

Source/SIMPLib/REST/V1Controllers/NamesOfFiltersController.cpp
```diff
@@ -0,0 +1,93 @@
...
+void NamesOfFiltersController::service(HttpRequest& request, HttpResponse& response)
+{
+
+  QString content_type = request.getHeader(QByteArray("content-type"));
+
+  QJsonObject rootObj;
+
+  response.setHeader("Content-Type", "application/json");
+
+  if(content_type.compare("application/json") != 0)
```
mmarineBlueQuartz: QString's == and != operators are faster than `.compare(const QString&) != 0` and easier to read.  If you do not need to ignore case or sort strings, anything that will be called frequently should rely on those operators.  If you need sorting information, then you should use `compare`.

Source/SIMPLib/REST/V1Controllers/SIMPLStaticFileController.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// -----------------------------------------------------------------------------
+//
+// -----------------------------------------------------------------------------
+void SIMPLStaticFileController::CreateInstance(ServerSettings* settings, QObject* parent)
+{
+  if(m_Instance != nullptr)
+  {
+    delete m_Instance;
```
mmarineBlueQuartz: `delete nullptr;` is valid C++.  You do not need the nullptr check.

Source/SVWidgetsLib/QtSupport/QtSDocServer.h
```diff
@@ -85,6 +89,8 @@ class SVWidgetsLib_EXPORT QtSDocServer : public QObject
 private:
   static QtSDocServer* self;
 
+  std::shared_ptr<ServerSettings> m_ServerSettings;
```
mmarineBlueQuartz, line 95: This is the ServerSettings pointer that should be used, but the .cpp file declares and uses s_ServerSettings as a raw pointer that does not belong to any object.  m_ServerSettings should be used instead.

Source/SVWidgetsLib/QtSupport/QtSDocServer.cpp
```diff
@@ -47,10 +47,10 @@
 QtSDocServer* QtSDocServer::self = nullptr;
 
 static HttpListener* httpListener = nullptr;
-static QSettings* sessionSettings = nullptr;
 static HttpSessionStore* sessionStore = nullptr;
-static QSettings* fileSettings = nullptr;
-static QSettings* listenerSettings = nullptr;
+
+static ServerSettings* s_ServerSettings = nullptr;
```
mmarineBlueQuartz: Raw global pointers should be avoided unless there is a very good reason to use them.  Also, `m_ServerSettings` was already declared in the header.  `s_ServerSettings` is redundant and lacks any form of memory management (no delete statement and not a std::shared_ptr).

Source/SVWidgetsLib/QtSupport/httpserver/QtSStaticFileController.h
```diff
@@ -5,12 +5,14 @@
 
 #pragma once
 
+#include <QCache>
+#include <QMutex>
```
mmarineBlueQuartz: QCache and QMutex are both part of QtCore

mmarineBlueQuartz (changes requested):
* The company name must be spelled correctly in copyright declarations.
* QtSDocServer has two different ServerSettings pointers.  Delete s_ServerSettings from the .cpp file and update all uses to the more secure m_ServerSettings.

Review range-based for loops for the future.  They're shorter to write, easier to read, and easier for the compiler to optimize.  There is no need to update them in this pull request, but they're great for iterating over containers, which you seemed to do a lot here.

Source/SIMPLib/REST/V1Controllers/SIMPLStaticFileController.cpp
joeykleingers: Really?  Did not know that.  Will keep that in mind in the future.  I assumed `delete nullptr;` would result in a crash.

mmarineBlueQuartz:
I also second the request for developer documentation for methods and classes.  The harder it is to maintain, the less it will be.

Source/SIMPLib/REST/V1Controllers/SIMPLStaticFileController.cpp
mmarineBlueQuartz: Before C++14, no destructors were called on nullptr, but it was implementation defined if the deallocation functions were called (which were already guaranteed to do nothing when given null pointers).  C++14 simplified this to never call deallocation functions on null pointers.

Source/SVWidgetsLib/QtSupport/QtSDocServer.h
imikejackson: Is the ServerSettings a static instance for this code? I think that was the intent

Source/SIMPLib/REST/V1Controllers/ExecutePipelineController.h
imikejackson: Done

Source/SVWidgetsLib/QtSupport/QtSDocServer.h
mmarineBlueQuartz: It is static without belonging to any object type and nothing cleans it up.  m_ServerSettings is never used or initialized.  If the variable belongs to the instance, memory becomes easier to manage and create "clean" copies if required.

Either keep s_ServerSettings or keep m_ServerSettings.  Having both when only one is used declares an intent that is completely inaccurate.  As m_ServerSettings is the more flexible and remains consistent with OOP, I am suggesting we keep that one.

Source/SIMPLib/REST/V1Controllers/NamesOfFiltersController.cpp
imikejackson: Done

Source/SIMPLib/REST/V1Controllers/SIMPLStaticFileController.cpp
imikejackson: Done

Source/SVWidgetsLib/QtSupport/QtSDocServer.cpp
imikejackson: Done and fixed

Source/SVWidgetsLib/QtSupport/QtSDocServer.h
imikejackson: Done and fixed

Source/SVWidgetsLib/QtSupport/httpserver/QtSStaticFileController.h
imikejackson: Done and fixed

imikejackson:
I think the issues raised have been addressed except for the range-based for loops which will just have to wait until the next round of funding.

imikejackson (approved):

mmarineBlueQuartz (approved):
This looks good to me.  The important parts have been fixed, and any other comments I have are just things to think about for the future.

PR #289, opened 2018-12-31
Use a data structure similar to ITK when storing 2D Tile information.

imikejackson:
Ours differs in that we store the row & column indices in addition to the raw
XY coordinates of each tile.


joeykleingers (approved):
Looks good.

PR #293, opened 2019-01-07
Fix crash when trying to read a simple dataset from HDF5

imikejackson:


mmarineBlueQuartz (approved):
This looks good.

PR #294, opened 2019-01-08
Fix a crash inside of dataCheck() in ArrayCalculator filter

imikejackson:
The crash can happen when a requested AttributeMatrix is NOT found which will
return a nullptr but is still used because the validity of the pointer is not
checked after return.


mmarineBlueQuartz (approved):
Looks good to me.

PR #295, opened 2019-01-09
Remove the notification at the end of the execute() method of filters.

imikejackson:
The FilterPipeline class now sends a PipelineMessage object at the completion of
each filter in a consistent form.


imikejackson:
EVERY other repository was hit with this update. If you agree with this PR then we can process the other 40 repositories using a local script.

mmarineBlueQuartz (approved):
This looks good to me.  Go ahead and update the repos.

PR #297, opened 2019-01-17
Feature/python wrapping

imikejackson:


Source/SIMPLib/DataContainers/AttributeMatrixProxy.cpp
```diff
@@ -285,3 +285,64 @@ void AttributeMatrixProxy::updatePath(DataArrayPath::RenameType renamePath)
     dataArrays.insert(newPath.getDataArrayName(), daProxy);
   }
 }
+
+// -----------------------------------------------------------------------------
+//
+// -----------------------------------------------------------------------------
+DataArrayProxy& AttributeMatrixProxy::getDataArrayProxy(const QString& name)
+{
+	for (QMap<QString, DataArrayProxy>::iterator iter = dataArrays.begin(); iter != dataArrays.end(); ++iter) // DataArray Level
```
mmarineBlueQuartz: Check your indentation for this entire section.  Either you are using tabs or a very different tab-to-spaces ratio than everywhere else.

mmarineBlueQuartz: There's also no newline at the end of the file, which we have everywhere else.  I'm not sure if we care about consistency for that, but apparently GitHub makes note of it.

Source/SIMPLib/DataContainers/AttributeMatrixProxy.h
```diff
@@ -137,6 +149,58 @@ class SIMPLib_EXPORT AttributeMatrixProxy
      */
     void updatePath(DataArrayPath::RenameType renamePath);
 
+
+	/**
+	 * @brief Get the attribute matrices (Python Binding)
+	 */
+	QMap<QString, DataArrayProxy> getdataArrays();
```
mmarineBlueQuartz: This chunk of new code is indented an additional time.

Source/SIMPLib/DataContainers/DataArrayProxy.cpp
```diff
@@ -324,3 +324,13 @@ void DataArrayProxy::updatePath(DataArrayPath::RenameType renamePath)
...
 }
+
+uint8_t DataArrayProxy::getflag()
+{
+	return flag;
+}
+
+void DataArrayProxy::setflag(uint8_t newFlag)
+{
+	flag = newFlag;
```
mmarineBlueQuartz: Both flag methods need an indentation check, and there's no new line at the end (again, not sure if we care about that)

Source/SIMPLib/DataContainers/DataContainerArrayProxy.cpp
```diff
@@ -544,3 +544,14 @@ QMap<QString, DataContainerProxy> DataContainerArrayProxy::readMap(QJsonArray js
...
+
+QMap<QString, DataContainerProxy> DataContainerArrayProxy::getdataContainers()
+{
+	return dataContainers;
+}
+
+
+ void DataContainerArrayProxy::setdataContainers(QMap<QString, DataContainerProxy> newDataContainers)
+{
+	 dataContainers = newDataContainers;
```
mmarineBlueQuartz: Indentation for goth dataContainer methods. 
Check capitalization on `getdataContainers` and `setdataContainers`.  If that is intentional, it should be marked in both .h and .cpp because it does not look like it is and invites anyone working on the file to "fix" it.

Source/SIMPLib/DataContainers/DataContainerArrayProxy.h
```diff
@@ -121,6 +126,17 @@ class SIMPLib_EXPORT DataContainerArrayProxy
...
+	/**
+	 * @brief Get the data containers (Python Binding)
+	 */
+	QMap<QString, DataContainerProxy> getdataContainers();
+
+	/**
+	 * @brief Get the data containers (Python Binding)
+	 * @param new dataContainers map
+	 */
+	 void setdataContainers(QMap<QString, DataContainerProxy>);
```
mmarineBlueQuartz: Indentation

Source/SIMPLib/DataContainers/DataContainerProxy.cpp
```diff
@@ -235,3 +235,64 @@ void DataContainerProxy::setFlags(uint8_t flag, AttributeMatrixProxy::AMTypeFlag
     }
   }
 }
+
+QMap<QString, AttributeMatrixProxy> DataContainerProxy::getattributeMatricies()
+{
+	return attributeMatricies;
```
mmarineBlueQuartz: This entire chunk needs to be checked for indentation.
Are these methods supposed to be lowercase until the third word, because I still want to "fix" that.  I'm starting to think I should not.

mmarineBlueQuartz: Not sure if we care, but there's no new line at the end of the file.

Source/SIMPLib/DataContainers/DataContainerProxy.h
```diff
@@ -127,6 +140,52 @@ class SIMPLib_EXPORT DataContainerProxy
     */
     bool readJson(QJsonObject& json);
 
+
+	/**
+	 * @brief Get the attribute matrices (Python Binding)
+	 */
+	QMap<QString, AttributeMatrixProxy> getattributeMatricies();
```
mmarineBlueQuartz: Indentation

Source/SIMPLib/Filtering/AbstractComparison.h
```diff
@@ -47,6 +47,8 @@
 */
 class SIMPLib_EXPORT AbstractComparison
 {
+	PYB11_CREATE_BINDINGS(AbstractComparison)
+	PYB11_PROPERTY(int UnionOperator READ getUnionOperator WRITE setUnionOperator)
```
mmarineBlueQuartz: Indentation

Source/SIMPLib/StatsData/PrecipitateStatsData.h
```diff
@@ -83,6 +83,21 @@ typedef QVector<FloatArrayType::Pointer> VectorOfFloatArray;
...
+	PYB11_PROPERTY(float MaxFeatureDiameter READ getMaxFeatureDiameter WRITE setMaxFeatureDiameter)
+	PYB11_PROPERTY(float MinFeatureDiameter READ getMinFeatureDiameter WRITE setMinFeatureDiameter)
+	PYB11_PROPERTY(VectorOfFloatArray FeatureSizeDistribution READ getFeatureSizeDistribution WRITE setFeatureSizeDistribution)
+	PYB11_PROPERTY(uint32_t FeatureSize_DistType READ getFeatureSize_DistType WRITE setFeatureSize_DistType)
+	PYB11_METHOD(PhaseType::Type getPhaseType)
+	PYB11_METHOD(FloatArrayType::Pointer generateBinNumbers)
+	PYB11_PROPERTY(float PhaseFraction READ getPhaseFraction WRITE setPhaseFraction)
+	PYB11_PROPERTY(float BoundaryArea READ getBoundaryArea WRITE setBoundaryArea)
+	PYB11_PROPERTY(float PrecipBoundaryFraction READ getPrecipBoundaryFraction WRITE setPrecipBoundaryFraction)
+	PYB11_METHOD(QString getStatsType)
```
mmarineBlueQuartz: Indentation

Source/SIMPLib/StatsData/PrimaryStatsData.h
```diff
@@ -82,6 +82,19 @@ typedef QVector<FloatArrayType::Pointer> VectorOfFloatArray;
...
+
+	PYB11_PROPERTY(float BinStepSize READ getBinStepSize WRITE setBinStepSize)
+	PYB11_PROPERTY(float MaxFeatureDiameter READ getMaxFeatureDiameter WRITE setMaxFeatureDiameter)
+	PYB11_PROPERTY(float MinFeatureDiameter READ getMinFeatureDiameter WRITE setMinFeatureDiameter)
+	PYB11_PROPERTY(VectorOfFloatArray FeatureSizeDistribution READ getFeatureSizeDistribution WRITE setFeatureSizeDistribution)
+	PYB11_PROPERTY(uint32_t FeatureSize_DistType READ getFeatureSize_DistType WRITE setFeatureSize_DistType)
+	PYB11_METHOD(PhaseType::Type getPhaseType)
+	PYB11_METHOD(FloatArrayType::Pointer generateBinNumbers)
+	PYB11_PROPERTY(float PhaseFraction READ getPhaseFraction WRITE setPhaseFraction)
+	PYB11_PROPERTY(float BoundaryArea READ getBoundaryArea WRITE setBoundaryArea)
```
mmarineBlueQuartz: Indentation

Source/SIMPLib/StatsData/StatsData.h
```diff
@@ -89,6 +89,11 @@ typedef QVector<FloatArrayType::Pointer> VectorOfFloatArray;
  */
 class SIMPLib_EXPORT StatsData
 {
+	PYB11_CREATE_BINDINGS(StatsData)
+	PYB11_METHOD(QString getStatsType)
+	PYB11_METHOD(PhaseType::Type getPhaseType)
```
mmarineBlueQuartz: Indentation

Wrapping/Python/Pybind11/CodeScraper/PythonBindingClass.cpp
```diff
@@ -432,7 +432,11 @@ QString PythonBindingClass::generateStaticNewCode()
         }
       }
       constructors << TAB << ".def(py::init([](" << s1 << ") {\n      return " << getClassName() << "::" << methodName << "(" << s2 << ");\n    }))" << NEWLINE_SIMPL;
-      constructors << TAB << ".def_static(\"" << methodName << "\", &" << getClassName() << "::" << methodName << ")" << NEWLINE_SIMPL;
+
+	  if (tokens.size() < 2 || tokens[1] != ::kOverload)
+	  {
+		  constructors << TAB << ".def_static(\"" << methodName << "\", &" << getClassName() << "::" << methodName << ")" << NEWLINE_SIMPL;
+	  }
```
mmarineBlueQuartz: Non-standard indentation.  Are tabs being used here?

Wrapping/Python/Pybind11/CodeScraper/PythonBindingClass.cpp
```diff
@@ -562,8 +566,17 @@ QString PythonBindingClass::generatePropertiesCode()
...
-         
+  
+  /* 
+    Replace any Python keywords in arguments
+     For now, just replace the occurence of 'lambda'
+  */
+  if (methodArgs.contains("lambda") || bodyCodes.contains("lambda"))
+  {
+	  methodArgs.replace("lambda", "lambda_1");
+	  bodyCodes.replace("lambda", "lambda_1");
```
mmarineBlueQuartz: Indentation

Wrapping/Python/Pybind11/CodeScraper/PythonBindingClass.cpp
```diff
@@ -620,16 +633,26 @@ QString PythonBindingClass::generateMethodCode()
...
       tokens.pop_back();
     }
+	if (tokens.contains(::kReturnValuePolicy))
+	{
+		hasReturnValuePolicy = true;
+		int index = tokens.indexOf(::kReturnValuePolicy);
+		returnValuePolicy = ", " + tokens[index + 1];
+		tokens.removeAt(index);
+		tokens.removeAt(index);
+	}
```
mmarineBlueQuartz: Indentation

Wrapping/Python/Pybind11/CodeScraper/PythonBindingsModule.cpp
```diff
@@ -164,6 +166,7 @@ void PythonBindingsModule::addDependency(QString superClassName, QString classNa
   {
     QObject* obj = new QObject(nullptr);
     obj->setObjectName(className);
+	obj->setProperty("superClass", superClassName);
```
mmarineBlueQuartz: Indentation.  Lines 155, 160, and 169

Wrapping/Python/Pybind11/CodeScraper/PythonBindingsModule.cpp
```diff
@@ -359,8 +358,11 @@ void PythonBindingsModule::generatePythonTestFile(const QString& outputPath, con
   
   for(auto object : m_ClassVector)
   {
-    out << "\n";
-    dumpRecursivePythonCode(0, object, out);
+	  if (object->property("superClass").toString().compare("AbstractFilter") == 0)
+	  {
+		  out << "\n";
+		  dumpRecursivePythonCode(0, object, out);
+	  }
```
mmarineBlueQuartz: Indentation

Wrapping/Python/Pybind11/CodeScraper/PythonBindingsModule.cpp
```diff
@@ -417,9 +422,12 @@ void PythonBindingsModule::generatePythonicInterface(const QString& outputPath,
     {
       continue;
     }
+	else if (bodyCodes.isEmpty()) {
+		continue;
+	}
```
mmarineBlueQuartz: Indentation

Wrapping/Python/Pybind11/Templates/Pybind11CustomTypeCasts.in.h
```diff
@@ -0,0 +1,620 @@
...
+					}
+
+					floatArrays.append(floatArray);
+				}
+				return floatArrays;
+			}
+		};
+
+	} // namespace detail
+} // namespace pybind11
```
mmarineBlueQuartz, line 627: Not sure if we care about .in files, but I think you're using tab

mmarineBlueQuartz (changes requested):
There are definitely a lot of places indentation gets thrown off.  Some of it is indentation from macros, and some of it looks like you are using tabs instead of spaces.

A bigger concern of mine is the non-standard lowercase method names like `getdataContainers`.  I started to think that was intentional, but then I found more PyBind methods that were added that followed our standard naming (i.e. along the lines of `getDataContainers`).  I am now confused on what's what, why we're doing it that way sometimes and not others, and if there's a specific reason, can we please mark it each time in the `@brief` and in the .cpp so no one breaks the python binding by "fixing" it?  Because that's probably going to be me.  That, and being inconsistent in our Python API is probably not a good thing either.

I did not review the Python files.  I know I am not the right person for that.

Also, the three line comment marker we put before methods in .cpp files is missing from many of the new methods but not all.

imikejackson:
the lowercase method names were updated appropriately. Those were non-conforming and not caught during development.

The indentations are getting taken care of. This might be a good time to just run clang format over the code base.

Source/SIMPLib/DataArrays/DataArray.hpp
```diff
@@ -855,6 +855,31 @@ class DataArray : public IDataArray
...
+	 */
+	void setArray(std::list<T> newArray)
+	{
+		if (newArray.size() != m_Size) {
+			return;
+		}
+		int i = 0;
+		for (auto elem : newArray)
+		{
+			m_Array[i++] = elem;
```
mmarineBlueQuartz: Clang format getArray and setArray

Source/SIMPLib/DataContainers/AttributeMatrixProxy.cpp
```diff
@@ -285,3 +278,63 @@ void AttributeMatrixProxy::updatePath(DataArrayPath::RenameType renamePath)
...
+
+  DataArrayProxy proxy;
+  proxy.name = name;
+  dataArrays.insert(proxy.name, proxy);
+  return dataArrays[name];
+}
+
+QMap<QString, DataArrayProxy> AttributeMatrixProxy::getDataArrays()
+{
+	return dataArrays;
```
mmarineBlueQuartz: getDataArrays through setFlag all have extra indentation and are missing the comment block above their definitions unlike other methods.

Source/SIMPLib/DataContainers/DataArrayProxy.cpp
```diff
@@ -324,3 +324,23 @@ void DataArrayProxy::updatePath(DataArrayPath::RenameType renamePath)
...
 }
+
+uint8_t DataArrayProxy::getFlag() const
+{
+	return flag;
+}
+
+void DataArrayProxy::setFlag(uint8_t newFlag)
+{
+	flag = newFlag;
```
mmarineBlueQuartz: Indentation is still off for getFlag and setFlag.  Comment blocks are missing above method definitions between getFlag and getName

Source/SIMPLib/DataContainers/DataContainer.h
```diff
@@ -73,8 +73,8 @@ class SIMPLib_EXPORT DataContainer : public Observable
   PYB11_METHOD(QString getInfoString ARGS InfoStringFormat)
   PYB11_METHOD(void addAttributeMatrix ARGS Name AttributeMatrix)
 
-  PYB11_METHOD(AttributeMatrix::Pointer getAttributeMatrix OVERLOAD const.QString.&, Name)
-  PYB11_METHOD(AttributeMatrix::Pointer getAttributeMatrix OVERLOAD const.DataArrayPath.&, Path)
+  PYB11_METHOD(AttributeMatrix::Pointer getAttributeMatrix OVERLOAD const.QString.&,Name)
+  PYB11_METHOD(AttributeMatrix::Pointer getAttributeMatrix OVERLOAD const.DataArrayPath.&,Path)
```
mmarineBlueQuartz, line 78: Is the space after the comma important at the end of this macro?

Source/SIMPLib/DataContainers/DataContainerArrayProxy.cpp
```diff
@@ -544,3 +544,13 @@ QMap<QString, DataContainerProxy> DataContainerArrayProxy::readMap(QJsonArray js
   }
   return map;
 }
+
+QMap<QString, DataContainerProxy> DataContainerArrayProxy::getDataContainers()
+{
+	return dataContainers;
```
mmarineBlueQuartz: getDataContainers and setDataContainers still require indentation checks.  The comment blocks above their definitions are also missing.

Source/SIMPLib/DataContainers/DataContainerProxy.cpp
```diff
@@ -228,10 +228,81 @@ void DataContainerProxy::setFlags(uint8_t flag, AttributeMatrixProxy::AMTypeFlag
...
+
+// -----------------------------------------------------------------------------
+//
+// -----------------------------------------------------------------------------
+AttributeMatrixProxy& DataContainerProxy::getAttributeMatrixProxy(const QString& name)
+{
+	for (QMap<QString, AttributeMatrixProxy>::iterator amIter = attributeMatricies.begin(); amIter != attributeMatricies.end(); ++amIter) // AttributeMatrix Level
+	{
+		AttributeMatrixProxy& am = amIter.value();
+    if(am.getName() == name)
```
mmarineBlueQuartz, line 336: An if statement inside a for loop should not be less indented than the for that contains it.

Source/SIMPLib/DataContainers/DataContainerProxy.cpp
mmarineBlueQuartz: getAttributeMatrices through getAttributeMatrixProxy still need indentation checks.  Also, none of them have the three-line comment blocks before their definitions.

Source/SIMPLib/DataContainers/DataContainerProxy.cpp
mmarineBlueQuartz: Also, `AttributeMatrixProxy proxy(name);` is oddly indented a few lines after this to match the for loop.

Source/SIMPLib/DataContainers/DataContainerProxy.h
```diff
@@ -35,141 +35,203 @@
...
+   * @brief Updates the proxy to match a renamed DataArrayPath
+   * @param renamePath
+   */
+  void updatePath(DataArrayPath::RenameType renamePath);
+
+  //----- Our variables, publicly available
+  uint8_t flag;
+  QString name;
+  unsigned int dcType;
+  QMap<QString, AttributeMatrixProxy> attributeMatricies;
```
mmarineBlueQuartz: We have getters and setters for all of these variables, and SIMPL code has been updated to use them.  Do we still want these to be public variables, or private?

Source/SIMPLib/DataArrays/DataArray.hpp
imikejackson: I'm going to clang-format the entire file.

Source/SIMPLib/DataContainers/DataContainer.h
imikejackson: YES. The space MUST be there for our parsing to work.

Source/SIMPLib/DataContainers/DataContainerProxy.h
imikejackson: Some of the codes still directly access them and I have not had the time to work through the ramifications of making them private (which I agree that we should do).

mmarineBlueQuartz (approved):
This looks good to me.

Source/SIMPLib/DataContainers/AttributeMatrix.h
```diff
@@ -96,7 +97,10 @@ class SIMPLib_EXPORT AttributeMatrix : public Observable
   PYB11_METHOD(bool doesAttributeArrayExist ARGS Name)
   PYB11_METHOD(IDataArray removeAttributeArray ARGS Name)
   PYB11_METHOD(int renameAttributeArray ARGS OldName NewName OverWrite)
-  PYB11_METHOD(IDataArray::Pointer getAttributeArray ARGS const.QString.&,name RETURN_VALUE_POLICY py::return_value_policy::reference)
+  PYB11_METHOD(IDataArray::Pointer getAttributeArray OVERLOAD const.QString.&,Name)
+  PYB11_METHOD(IDataArray::Pointer getAttributeArray OVERLOAD const.DataArrayPath.&,Path)
```
mmarineBlueQuartz, line 101: Missing a space between the comma and Name / Path

Source/SIMPLib/DataContainers/DataArrayProxy.h
```diff
@@ -171,28 +180,86 @@ class SIMPLib_EXPORT DataArrayProxy
    * @brief setName
    * @param name
    */
-  void setName(const QString& name);
+  void setName(const QString& m_Name);
```
mmarineBlueQuartz: Are you sure about this rename?

Source/SIMPLib/DataContainers/DataContainer.h
mmarineBlueQuartz: The space is missing again.

Source/SIMPLib/DataContainers/DataContainerProxy.h
```diff
@@ -193,7 +201,7 @@ class SIMPLib_EXPORT DataContainerProxy
    * @param the name of the attribute matrix
    * @return the attribute matrix proxy
    */
-  AttributeMatrixProxy& getAttributeMatrixProxy(const QString& name);
+  AttributeMatrixProxy& getAttributeMatrixProxy(const QString& m_Name);
```
mmarineBlueQuartz, line 207: Are you sure about renaming the parameters in setFlag and getAttributeMatrixProxy?  If not, you might want to update the insertAttributeMatrix declaration as well.

mmarineBlueQuartz: Same with the flag parameter in setFlags below.

Source/SIMPLib/Filtering/ComparisonSet.h
```diff
@@ -53,6 +54,7 @@ class SIMPLib_EXPORT ComparisonSet : public AbstractComparison
   PYB11_PROPERTY(int UnionOperator READ getUnionOperator WRITE setUnionOperator)
   PYB11_METHOD(void addComparison ARGS AbstractComparisonPtr,comparison)
   PYB11_METHOD(void insertComparison ARGS int,index AbstractComparisonPtr,comparison)
```
mmarineBlueQuartz, line 56: I am going to assume spaces are important here, too.  Did I miss this in the earlier commits, or is not having a space actually valid here?

Source/SIMPLib/Filtering/ComparisonSet.h
```diff
@@ -53,6 +54,7 @@ class SIMPLib_EXPORT ComparisonSet : public AbstractComparison
...
+  // clang-format om
```
mmarineBlueQuartz: `// clang-format on` not om

Source/SIMPLib/DataArrays/StringDataArray.h
```diff
@@ -63,6 +64,8 @@ class SIMPLib_EXPORT StringDataArray : public IDataArray
   PYB11_METHOD(void setValue ARGS size_t,i const.QString.&,value)
```
mmarineBlueQuartz, line 64: Are additional spaces necessary here?

Source/SIMPLib/Filtering/ComparisonInputs.h
```diff
@@ -97,18 +97,16 @@ typedef struct
...
-
+  // clang-format off
   PYB11_CREATE_BINDINGS(ComparisonInputs)
-
   PYB11_CREATION()  
-
   PYB11_PROPERTY(QVector<ComparisonInput_t> Inputs READ getInputs)
-
   PYB11_METHOD(void addInput OVERLOAD const.QString,dataContainerName const.QString,attributeMatrixName const.QString,arrayName int,compOperator double,compValue)
   PYB11_METHOD(void addInput OVERLOAD const.ComparisonInput_t.&,input)
```
mmarineBlueQuartz, line 105: I believe you need a space before `input`.

mmarineBlueQuartz (changes requested):
Minimum Requirement:

- Fix one `// clang-format om`

Strongly Suggested:

- Put const back in methods that removed it from proxy parameters.  Within those methods, please use them for each range-based for loop.  Obviously, not of these cases make sense to be const, but the amount of non-const references in places that should not change the value makes me worry about how easy it would be to break pipelines or UI.
   - Add const versions (const T& name() const;)  of each proxy getter that returns a reference so that const can be used where appropriate.  For instance, adding `const QMap<QString, DataArrayProxy>& getDataArrays() const;` to AttributeMatrixProxy in addition to the non-const version.  The lack of having these methods declared in both versions seems to be why const was removed from proxy parameters.
   - Logic behind this: If something should not be changed, make it a compiler error to do so.  This is why each of the STL containers have const and non-const versions defined for most getters.
   - If you do not want to duplicate code between const and non-const methods, an acceptable solution is to call the const version from non-const method and cast away const on the return statement.  It is not pretty, but it is safe because the object itself (and thus its variables) are not promised to be constant when calling a non-const method.

Edit: Apparently space requirements in PYB11_METHOD have changed.

Source/SIMPLib/DataContainers/DataContainer.h
imikejackson: Actually, no the space should NOT be there. I am working a better solution to this.

Source/SIMPLib/Filtering/ComparisonInputs.h
imikejackson: Nope.

Source/SIMPLib/DataArrays/StringDataArray.h
imikejackson: Yes

imikejackson:
In some places we do need to edit the underlying storage. Maybe we should then just return an iterator that allows outside codes to modify the underlying storage? That was the idea behind add the "insert*()" methods into the proxy classes.

mmarineBlueQuartz:
Iterators would be a good idea as long as we have both iterator and const_iterator versions (QMap already offers both). It would be a more convenient way to simply loop over a proxy without having to create additional variables.

Source/SIMPLib/DataContainers/DataArrayProxy.h
imikejackson: Fixed

PR #299, opened 2019-01-23
Updating ReadASCIIData filter to support relative paths.

joeykleingers:
Updating ReadASCIIData filter and its widget to support relative paths.  Added a new SVBoldLabel control widget so that the selected path is bolded, and made the new absolute file path that appears underneath inherit from SVSmallLabel.

Signed-off-by: Joey Kleingers <joey.kleingers@bluequartz.net>


imikejackson (changes requested):
Remove the BoldLabel class and use the SVSmallLabel class.

imikejackson (approved):

PR #301, opened 2019-01-30
Fix type of cast, check for valid pointers before using.

imikejackson:
SVPipelineView needs sanity checks on pointers before using them because the source
of the variable could return null pointers.

### DIFF
--- a/Source/SVWidgetsLib/Widgets/SVPipelineView.cpp
+++ b/Source/SVWidgetsLib/Widgets/SVPipelineView.cpp
@@ -329,6 +329,16 @@ void SVPipelineView::preflightPipeline()
   emit clearIssuesTriggered();
 
   PipelineModel* model = getPipelineModel();
+  if(nullptr == model)
+  {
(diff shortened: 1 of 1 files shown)


imikejackson:
Not sure we really need that big output block to StdOut.

mmarineBlueQuartz (approved):
This looks good.

mmarineBlueQuartz (changes requested):
Upon further inspection, lines 383 and 388 also have the same nullptr error.  PipelineModel does not clear items when a filter is removed from the pipeline, and thus model->filter(childIndex) will return nullptr for legitimate indices for the pipeline.  The initial check if the PipelineModel is null is valid as there is no guarantee that the view's QAbstractItemModel* has to be castable to PipelineModel* other than that a failure to do so should not happen.  The same was assumed for `model->filter(childIndex);` not returning null in multiple places.

imikejackson:
This PR has been supplanted by #304

PR #303, opened 2019-01-30
Topic/302 import ascii fixes

imikejackson:


mmarineBlueQuartz (commented):
Some of the non-const reference parameters should be const when the object in question is not being changed.  I understand that cases where we call methods on objects like DataContainerArray::Pointer requires it to be non-const, but I don't think using a const value to initialize another is a compile error in this situation.  For example, DataFormatPage's constructor takes a DataContainerArray::Pointer& to initialize `, m_Dca(dca)`.

I missed this on earlier reviews, but in SVPipelineView::preflightPipeline(), the first section where the model is being iterated over with setData(childIndex...) and we check if the filter is enabled, we should not be iterating over the same range as the pipeline.  In cases where a PipelineItem contains invalid data, the model does not have the same count as the pipeline, and not all values will be cleared.  The same is true in the last for loop where we should be using the model's rowCount() for the appropriate pipeline for the same reason.  If we want to leave this as a temporary fix and simply redesign the model instead, that's fine.  I just wanted to make sure we don't miss this point.

imikejackson:
The SVPipelineView stuff should not be in here, let me rebase so these topics are again separate.

PR #304, opened 2019-01-31
SVPipelineView additional checks for nullptr and/or valid objects.

imikejackson:

### DIFF
--- a/Source/SVWidgetsLib/Widgets/SVPipelineView.cpp
+++ b/Source/SVWidgetsLib/Widgets/SVPipelineView.cpp
@@ -1,7 +1,5 @@
-#include <utility>
-
 /* ============================================================================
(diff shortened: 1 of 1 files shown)


mmarineBlueQuartz (commented):
Copying comment from #303  
I missed this on earlier reviews, but in SVPipelineView::preflightPipeline(), the first section where the model is being iterated over with setData(childIndex...) and we check if the filter is enabled, we should not be iterating over the same range as the pipeline. In cases where a PipelineItem contains invalid data, the model does not have the same count as the pipeline, and not all values will be cleared. The same is true in the last for loop where we should be using the model's rowCount() for the appropriate pipeline for the same reason. If we want to leave this as a temporary fix and simply redesign the model instead, that's fine. I just wanted to make sure we don't miss this point.

imikejackson:
This is a temp fix until a complete redesign is implemented.

PR #305, opened 2019-01-31
Properly set the tuple dim values. Set the number of headers correctly

### LINKED ISSUE #302: Import ASCII Data bugs when importing data
User's report too many columns being required because a line in the header has a bunch of columns.

Tuple dimensions are not correctly populating the table. The values are being reset to a single dimension when there are 3 in the pipeline file.

imikejackson:
The headers were only allowed to grow so if there was a line in the header portion
of the file that had a lot of tokens, the number of columns of data would be set
to this value but never allowed to contract. The fix was to ensure an update to
the headers when the import line number was set.

The tuple dimensions were reset when reading from a pipeline file. The initial read
of the json values was correct but once placed in the table they were replaced/reset
with default number of rows further down in the initialization.

Updated some of the API to use a const QString& reference instead of making copies.

Implemented a caching algorithm for the first 50 lines of the text file and put a
QFileSystemWatcher on the file. If the file changes the cached lines will also be
updated.

updates #302. Fixes #302.


imikejackson:
This is a new PR to replace the #303 that was messed up and closed.

Source/SVWidgetsLib/Widgets/ImportASCIIDataWizard/DataFormatPage.cpp
```diff
@@ -698,7 +705,12 @@ void DataFormatPage::on_startRowSpin_valueChanged(int value)
   bool spaceAsDelimiter = field("spaceAsDelimiter").toBool();
   bool consecutiveDelimiters = field("consecutiveDelimiters").toBool();
 
-  QStringList lines = ImportASCIIDataWizard::ReadLines(m_InputFilePath, value, ImportASCIIDataWizard::TotalPreviewLines);
+  // Get our lines from the file cache.
+  QStringList lines;
+  for(int i = value-1; i < m_PreviewLinesCache.size(); i++)
+  {
+    lines.push_back(m_PreviewLinesCache[i]);
+  }
```
mmarineBlueQuartz, line 719: Is there any way value can be 0?  This method does not have any guards against it, and QStringList does not define what behavior is for [-1], only that "_i_ must be a valid index position in the list".

Source/SVWidgetsLib/Widgets/ImportASCIIDataWizard/ImportASCIIDataWizard.cpp
```diff
@@ -300,12 +301,16 @@ void ImportASCIIDataWizard::InsertTokenizedLines(QList<QStringList> tokenizedLin
...
   int vHeaderIndex = firstRowHeaderIndex;
 
   for(int row = 0; row < tokenizedLines.size(); row++)
   {
     QStringList tokenizedLine = tokenizedLines[row];
-
+    if(tokenizedLine.size() > maxNumHeaders)
+    {
+      maxNumHeaders = tokenizedLine.size();
+    }
```
mmarineBlueQuartz: `maxNumHeaders = std::max(maxNumHeaders, tokenizedLine.size());`

Source/SVWidgetsLib/Widgets/ImportASCIIDataWizard/DataFormatPage.cpp
imikejackson: The value passed in _should_ be coming from a QSpinBox that has a minimum set to 1. So in theory value should always be at least 1. But things change and someone else might call that method with a zero.

Source/SVWidgetsLib/Widgets/ImportASCIIDataWizard/ImportASCIIDataWizard.cpp
```diff
@@ -275,11 +276,11 @@ QStringList ImportASCIIDataWizard::ReadLines(const QString& inputFilePath, int b
...
   {
     QTextStream in(&inputFile);
-
+    QString aLine;
     for(int i = 1; i < beginLine; i++)
     {
       // Skip all lines before "value"
-      in.readLine();
+      aLine = in.readLine();
     }
```
mmarineBlueQuartz, line 283: I understand why aLine was created, but it took some digging to figure out the assumption that the last value of aLine matters because we bother to waste memory on it was invalid.  Just use `(void)in.readLine();`  We already use `(void)x;` several places in the code as it is the definition of Q_UNUSED, and it's faster to read than trying to figure out why I care about aLine.

imikejackson: I think that was left over from debugging?

Source/SVWidgetsLib/Widgets/ImportASCIIDataWizard/ImportASCIIDataWizard.cpp
imikejackson: Done

mmarineBlueQuartz (changes requested):
I still think DataFormatPage constructor should take const refs instead of just references, but since they're both pointers and not used directly, I wouldn't worry about it.

`on_startRowSpin_valueChanged(int value)` does worry me because the code itself does not check for 0, and that would give undefined behavior.  I am apparently going with the unfounded assumption that it can't be negative, but given `startRow` is in the name, that sounds safe.  Is this using 0-based or 1-based indexing?

Other than that, I only have two readability suggestions noted.

mmarineBlueQuartz (approved):
Apparently, I didn't have time to finish writing the earlier review before they were fixed.

imikejackson:
Clang-Tidy says either use const ref in the constructor or use std::move() on the variables. It still seems silly to pass a shared_ptr by reference but I'm ok with moving up to a more C++11/14 standards conforming code base.

imikejackson:
The user will see  1 based indexing in the UI much like a spread sheet. The QSpinBox has a minimum of 1. THis should be ok but I put in some defensive code to ensure that if value < 1 we just bail. Silently.

mmarineBlueQuartz:
I can understand both const ref and std::move(), but leaving them as only references felt very odd to read.  I did find [this](https://stackoverflow.com/a/8741626) answer / comment mentioning correct shared_ptr usage, although it might be outdated by now.  I'd go with what Clang suggests, but the links might be able to better describe why.  I think I saw this several months ago, but I could use a refresher on it as well.

imikejackson:
Watching the video and reading some more I will go back and try to update them to const reference and see what break....

Issue #300, opened 2019-01-29
PipelineRunner does not search in lib directory for plugins on Linux


imikejackson:
The .plugin files are installed into "lib" directory on Linux which is not search by default for PipelineRunner application. This will cause a basic failure to load any filters except for core filters.

The work around until the next release is for the user to set the SIMPL_PLUGIN_PATH environment variable to point to the "lib" directory inside the DREAM.3D installation folder.

PR #306, opened 2019-01-31
Add the 'lib' directory to the search paths on Linux builds

imikejackson:


imikejackson:
updates #300

PR #307, opened 2019-02-04
Fix issue where the parent of a stored path was used instead of the path

imikejackson:
Bug was discovered working with various widgets that save a path instead of a file path.

API change to make it more consistent as to what is being set/get


mmarineBlueQuartz (approved):
This looks good to me.  It reads a lot more easily than before.

PR #310, opened 2019-02-07
Improvements to the Write ASCII Data filter. Single file mode

### LINKED ISSUE #309: Export ASCSII Data needs to have an option to export a single file.
This is transferred from https://github.com/BlueQuartzSoftware/DREAM3D/issues/878

imikejackson:
Added the ability for the filter to write a single file with all the data arranged
in a column format. The original mode is called 'Multi File Export' and is the
default.
Documentation is updated.
Unit Test Updated.
DREAM.3D updated to use this filter in an Example Pipeline (INL Export).

Updates #309. Fixes #309

### DIFF
--- a/Source/SIMPLib/Documentation/SIMPLibFilters/CoreFilters/WriteASCIIData.md
+++ b/Source/SIMPLib/Documentation/SIMPLibFilters/CoreFilters/WriteASCIIData.md
@@ -1,5 +1,4 @@
-Export ASCII Data 
-=============
+# Export ASCII Data #
 
 ## Group (Subgroup) ##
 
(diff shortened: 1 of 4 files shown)


Source/SIMPLib/Documentation/SIMPLibFilters/CoreFilters/WriteASCIIData.md
```diff
@@ -49,11 +50,13 @@ Choice of delimiter is as follows:
 ## Parameters ##
 
 | Name             | Type | Description |
-|------------------|------|------------|
-| Output Path | File Path | The output file path |
-| File Extension | String | File extension for output file(s) |
+|------------------|------|---------_---|
```
mmarineBlueQuartz: I don't think `_` is what you meant to put here.

mmarineBlueQuartz (approved):
I don't see anything particularly wrong, but there were two things that stood out as odd to me.
You have static const values in a nameless namespace in a .cpp file.  I don't think static does anything here, so it seems odd that it is.
Second, in dataCheck(), you reinvented a switch block with if / else using those constants.  This feels like a case where an enum and switch should have been used instead.  It doesn't need to go so far as to be an enum class, but I feel like this would both read and declare intent better as it already follows a flow of
`case const_1:`
` ... break;`
`case const_2:`
` ... break;`
`default:`
` ... break;`

Source/SIMPLib/Documentation/SIMPLibFilters/CoreFilters/WriteASCIIData.md
imikejackson: Done

PR #311, opened 2019-02-07
Updates to ThresholdObjects2 related to styling and layout

imikejackson:
Updated Human Label to 'Threshold Objects (Advanced)'
Removed hard coded styling since it is incompatible with the StyleSheet.
Added SVFrame to SVControlWidgets
Adjusting layouts of several widgets to make them more pleasing visually.

These fixes enable fixes in the Branding repository to look better.


mmarineBlueQuartz (commented):
This seems to work, but it would be a good idea to find a way to get the border styling back in for the selected ComparisonContainerWidget.  I agree that, especially with global styles, setStyleSheet is no longer a viable solution, so piggybacking on either the object name or a Boolean selected property would be good.  If going down that road, however, neither setting the name nor a property causes the styling to be updated on itself.  `style()->unpolish(this); style()->polish(this);` will be required in each location where `setStyleSheet(QString)` had been called.

imikejackson:
I just used the SVStyle::Instance to get the color the filter highlight instead and adjusted the style a bit. Also fixed a bug when moving a ComparisonSet would reset the selection.

PR #312, opened 2019-02-08
Enhanced the error message returned when json parsing errors happen.

### LINKED ISSUE #308: Provide better feedback to users when parsing JSON fails
Current techniques required for batch processing ask the user to use an existing DREAM.3D pipeline as a "template", perform replacements within the file with keywords that the user will then replace using another language. If they try to load this "template" pipeline file into DREAM.3D they get a cryptic error message.
 A concrete example would be to replace what should be an integer value inside the JSON file with a keyword such as @REPLACE_ME@. The JSON parsing will fail.

imikejackson:
Added more text to the error message based on Qt5's message.
Adjusted the SVStyle to replace newline characters with <br> when creating HTML
wrapped messages.
Return early in SVPipelineView when trying to read a pipeline file that fails.

updates #308. fixes #308.


joeykleingers (approved):
Looks good.  Merge it in.

PR #314, opened 2019-02-09
Mitigate crashes during preflight by disconnecting signals/slots.

imikejackson:
Also use an unlimited undo stack size.

The AddFilterCommand or the RemoveFilterCommand are part of the UndoStack and are created every time a filter is created or removed from the pipeline. The signal/slot was hooked up using a lambda as the target of the slot. The QUndoStack max undo size was set to 10. After the 10th filter was added to the pipeline, attempting to set a value in the first filter input widget, which would cause a preflight would crash the pipeline. This is because when the 11th filter is added the first "AddFilterCommand" object gets removed from the stack and destroyed. The signal/slots were never disconnected before the object is destroyed. This repercussion is that a signal/slot connection is still going to be executed on a piece of memory will get corrupted by being written over with new allocations as part of the program.

The implemented solution is 2 fold: During the destructor of the AddFilterCommand and RemoveFilterCommand we call the disconnectSignalsSlots() function. We also use a QMetaObject::Connection object to store the connection to the lambda and then disconnect the QMetaObject::Connection during the destructor. We also now use an unlimited undo stack.

### DIFF
--- a/Source/SVWidgetsLib/Widgets/util/RemoveFilterCommand.cpp
+++ b/Source/SVWidgetsLib/Widgets/util/RemoveFilterCommand.cpp
@@ -92,7 +92,15 @@ RemoveFilterCommand::RemoveFilterCommand(std::vector<AbstractFilter::Pointer> fi
 // -----------------------------------------------------------------------------
 //
 // -----------------------------------------------------------------------------
-RemoveFilterCommand::~RemoveFilterCommand() = default;
+RemoveFilterCommand::~RemoveFilterCommand()
+{
+  for(const auto& filter : m_Filters)
+  {
+ //   qDebug() << "~RemoveFilterCommand(): ";
+ //   if(nullptr != filter.get()) { qDebug() << filter->getNameOfClass(); }
+    disconnectFilterSignalsSlots(filter);
+  }
+}
 
 // -----------------------------------------------------------------------------
 //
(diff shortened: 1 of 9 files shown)


Source/SVWidgetsLib/Widgets/util/RemoveFilterCommand.cpp
```diff
@@ -277,16 +285,16 @@ void RemoveFilterCommand::connectFilterSignalsSlots(AbstractFilter::Pointer filt
...
+  //PipelineModel* model = m_PipelineView->getPipelineModel();
 
   QObject::disconnect(filter.get(), &AbstractFilter::filterCompleted, nullptr, nullptr);
 
   QObject::disconnect(filter.get(), &AbstractFilter::filterInProgress, nullptr, nullptr);
-
-  FilterInputWidget* fiw = model->filterInputWidget(index);
-
-  QObject::disconnect(fiw, &FilterInputWidget::filterParametersChanged, nullptr, nullptr);
+  QObject::disconnect( m_connection );
```
mmarineBlueQuartz, line 299: RemoveFilterCommand never actually sets m_connection (this is the only time the variable is used in the entire .cpp file), and because the class operates on many filters, this variable should be a vector of QMetaObject::Connections with one per filter instead of a single entity that will be overwritten by each call to connectFilterSignalsSlots(const AbstractFilter::Pointer&).

Issue #316, opened 2019-02-10
Import HDF5 File does not check for duplicate DataArrayPath entries.


imikejackson:
recreate by adding a pair of Import HDF5 Data filters into the pipeline and making the same selections in each filter. The second instance of the filter *should* be complaining about there is already a data set at the path, but does not.

imikejackson:
Check the PR for an initial implementation and see if you agree with it.

PR #317, opened 2019-02-10
Implement proper DataArrayPath checking during preflight.

imikejackson:
The ImportHDF5DatasetWidget was also updated to be consistent with the features
of the AbstractIOWidget where we can drag-and-drop files, show absolute paths, and
reveal the file in the operating system's file navigation(macOS Find, Windows Explorer).

AbstractIOFileWidget also received a small update to fix a logic error when verifying
that a file exists.

updates #316


joeykleingers (approved):
Looks good to me.  Merge it in.

PR #319, opened 2019-02-11
WIP: Add an STL API to the DataArray class.

imikejackson:
Just some work down in the experimental area of SIMPL. There is a new AttributeArray
class and a small test program to act as a playground.

### DIFF
--- a/Source/Experimental/AttributeArrayIteratorExample/AttributeArray.hpp
+++ b/Source/Experimental/AttributeArrayIteratorExample/AttributeArray.hpp
@@ -1,0 +1,610 @@
+#pragma once
+
+#include <iostream>
+#include <string>
+#include <vector>
+#include <functional>
+#include <numeric>
+
+// -----------------------------------------------------------------------------
+template <typename T> class AttributeArray
+{
+  public:
+
+    using comp_dims_type = std::vector<size_t>;
+
+    using size_type = size_t;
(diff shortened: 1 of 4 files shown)


imikejackson:
Some of this is straight forward to implement. While I am not sure we can get 100% API compatibility with std::vector we can get pretty close.

I also started tweaking AttributeMatrix to expose the QMap<A,B> container that is used to hold the Attribute Arrays. There are some basic iterators and operators exposed as a start. We should probably change the internal container to be a std::map<> instead of a QMap<>. Should do this for all the classes really.

Source/Experimental/AttributeArrayIteratorExample/AttributeArray.hpp
```diff
@@ -0,0 +1,610 @@
...
+
+      private:
+        pointer ptr_;
+        size_t num_comps_;
+    };
+
+    class const_tuple_iterator
+    {
+      public:
+        using self_type = tuple_iterator;
```
mmarineBlueQuartz: I believe this should be `using self_type = const_tuple_iterator;`

imikejackson: yep.

imikejackson:
This is being superseded by another branch that is being worked on

PR #320, opened 2019-02-11
Properly assign and later cleanup the connection object.

imikejackson:
This takes into account that multiple filters can be assigned to a command.

### DIFF
--- a/Source/SVWidgetsLib/Widgets/util/AddFilterCommand.cpp
+++ b/Source/SVWidgetsLib/Widgets/util/AddFilterCommand.cpp
@@ -135,9 +135,9 @@ void AddFilterCommand::undo()
   for(int i = 0; i < m_Filters.size(); i++)
   {
     QPersistentModelIndex filterIndex = model->indexOfFilter(m_Filters[i].get());
-
     removeFilter(filterIndex);
   }
+  m_Connections.clear();
 
   QString statusMessage;
   if(m_Filters.size() > 1)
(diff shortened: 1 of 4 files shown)


Source/SVWidgetsLib/Widgets/util/AddFilterCommand.cpp
```diff
@@ -301,9 +302,13 @@ void AddFilterCommand::disconnectFilterSignalsSlots(const AbstractFilter::Pointe
...
- // PipelineModel* model = m_PipelineView->getPipelineModel();
 
   QObject::disconnect(filter.get(), &AbstractFilter::filterCompleted, nullptr, nullptr);
 
   QObject::disconnect(filter.get(), &AbstractFilter::filterInProgress, nullptr, nullptr);
+
+  for(auto& connection : m_Connections)
+  {
+    QObject::disconnect(connection);
+  }
```
mmarineBlueQuartz: This method is called per filter so having a map to the filter might make more sense here, but m_Connections needs to be cleared after disconnecting the connection.  Otherwise, m_Connections will continue to grow larger every time undo() redo() are called.

mmarineBlueQuartz: This is duplicated in RemoveFilterCommand.

mmarineBlueQuartz (approved):
Looks good to me.

imikejackson:
It works, but just by chance. I'll try a map<Filter, Connection> instead.

PR #322, opened 2019-02-15
[WIP] MultiSelection widget revamp

nathanblair:
Work done on MultiDataContainerSelection widget includes simplifying down to single list, using currently selected items for setting the filter parameter property instead. It also bases the widget class on a ListView instead of a ListWidget, which relies on using a QAbstractItemModel (specifically the QStringListModel), making it easier to track and update changes with much less custom code.

This could then be extended to the other Multi-*** select widgets as well.

This PR should stand as a reminder that this work is ongoing and could continue as/after work on blending and stitching is completed.

### DIFF
--- a/Source/SVWidgetsLib/FilterParameterWidgets/MultiDataContainerSelectionWidget.cpp
+++ b/Source/SVWidgetsLib/FilterParameterWidgets/MultiDataContainerSelectionWidget.cpp
@@ -57,6 +57,9 @@
 #include "FilterParameterWidgetUtils.hpp"
 #include "FilterParameterWidgetsDialogs.h"
 
+// -----------------------------------------------------------------------------
+// Constructor - calls Qt setupUi and setupGui
+// -----------------------------------------------------------------------------
 MultiDataContainerSelectionWidget::MultiDataContainerSelectionWidget(FilterParameter* parameter, AbstractFilter* filter, QWidget* parent)
 : FilterParameterWidget(parameter, filter, parent)
 {
@@ -67,8 +70,14 @@ MultiDataContainerSelectionWidget::MultiDataContainerSelectionWidget(FilterParam
   setupGui();
 }
 
+// -----------------------------------------------------------------------------
+// Default destructor
+// -----------------------------------------------------------------------------
 MultiDataContainerSelectionWidget::~MultiDataContainerSelectionWidget() = default;
 
+// -----------------------------------------------------------------------------
+// Connects signals and adds default values to select widget
+// -----------------------------------------------------------------------------
 void MultiDataContainerSelectionWidget::setupGui()
 {
   // Sanity Check the filter and the filter parameter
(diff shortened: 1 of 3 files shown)


Source/SIMPLib/ITK/itkFijiConfigurationFileReader.hpp
```diff
@@ -0,0 +1,127 @@
...
+          imageTileData.row = i;
+          imageTileData.col = j;
+
+          rowData.push_back(imageTileData);
+        }
+
+        fileData.push_back(rowData);
+      }
+
+      return fileData;
```
mmarineBlueQuartz: Unless a method or class is templated, constexpr, or inline, try to avoid putting method definitions in the header.  I'm aware that this is done in other places, so I'm not going to say it needs to be changed.  Just try to avoid it in the future.

This seems to be specific enough it won't matter much, but the reason to avoid it is to avoid forcing recompiles for every file that even indirectly includes this if even a single line of a method definition changes.  The reason templates, constexpr, and inline are excused are because they require definitions where they are declared for the sake of compiling.  Constexpr and inline are both pretty forceful runtime optimizations and require consideration when using in part for these same reasons.  Considering constexpr throws a fit with strings, two of the three reasons are already ruled out.  Inline is only useful for small methods like getErrorCode() and getErrorMessage() when they are called in tight loops or other time-critical places where runtime is slowed down due to the function call overhead.  Do note that this, like constexpr, is handled at compile time and will slow that down slightly in exchange for the runtime performance.

Source/SVWidgetsLib/FilterParameterWidgets/FilterParameterWidget.h
```diff
@@ -149,22 +149,6 @@ class SVWidgetsLib_EXPORT FilterParameterWidget : public QFrame
...
-     */
-    bool hasValidFilePath(const QString &filePath);
-
-    /**
-     * @brief verifyPathExists
-     * @param filePath
-     * @param lineEdit
-     * @return
-     */
-    bool verifyPathExists(QString filePath, QLineEdit* lineEdit);
```
mmarineBlueQuartz: Have we tested this with all of our plugins?  Removing methods from classes as abstract as FilterParameterWidget makes me worried about breaking builds due to expected methods no longer being available, even if the methods themselves probably should not have been available outside of specific instances of this class.

nathanblair: I believe this change was actually made in the ongoing feature/OSD_GUI branch being worked on. I don't expect this PR to be ready before that branch is submitted so I put a PR to merge into the develop because I believe the changes in the OSD_GUI branch will be in develop before this guy gets merged in, which should ease the merging a bit.

Source/SIMPLib/ITK/itkFijiConfigurationFileReader.hpp
nathanblair: This file I cannot speak to. Idk if @joeykleingers has anything to chime in and say? regardless, I don't think it has any bearing on the mutli-selection widget nature of the PR.

PR #324, opened 2019-02-19
Fixing compile error on GCC 4.8.5 in ColorToGrayScaleTest

imikejackson:


nathanblair:
I don't have a GCC to test this against at the moment but I'd assume this would be fine to change!

PR #325, opened 2019-02-26
Moving duplicate file-related functions to QtSFileUtils

joeykleingers:
Removes the various duplicate verifyPathExists and hasValidPath methods.  All plugins and widgets can then call the same methods from the QtSFileUtils class.


Source/SVWidgetsLib/FilterParameterWidgets/FilterParameterWidget.h
```diff
@@ -149,22 +149,6 @@ class SVWidgetsLib_EXPORT FilterParameterWidget : public QFrame
...
-     */
-    bool hasValidFilePath(const QString &filePath);
-
-    /**
-     * @brief verifyPathExists
-     * @param filePath
-     * @param lineEdit
-     * @return
-     */
-    bool verifyPathExists(QString filePath, QLineEdit* lineEdit);
```
imikejackson, line 165: the QString should be passed by const ref: const QString& filePath.

Source/SVWidgetsLib/QtSupport/QtSFileUtils.cpp
```diff
@@ -179,3 +181,99 @@ void QtSFileUtils::ShowPathInGui(QWidget* parent, const QString &pathIn)
   }
 }
 
+// -----------------------------------------------------------------------------
+//
+// -----------------------------------------------------------------------------
+bool QtSFileUtils::VerifyPathExists(QString filePath, QLineEdit* lineEdit)
+{
```
imikejackson, line 185: pass by const ref: const QString& filePath

Source/SVWidgetsLib/QtSupport/QtSFileUtils.h
```diff
@@ -58,6 +60,21 @@ class SVWidgetsLib_EXPORT QtSFileUtils
      */
     static void ShowPathInGui(QWidget* parent, const QString &path);
 
+    /**
+     * @brief verifyPathExists
+     * @param filePath
+     * @param lineEdit
+     * @return
+     */
+    static bool VerifyPathExists(QString filePath, QLineEdit* lineEdit);
```
imikejackson: const QString& filePath

imikejackson:
Please always pass QString by const ref  unless the method is actually manipulating that string object.

imikejackson:
And clang-format the changed code...

PR #326, opened 2019-02-27
Now uses intended ConvertColorToGrayScale filter

nathanblair:
Previous version was not the filter revised in the prior PR. This one works as expected and has been tested with the SIMPLUnitTest program with all tests passing.

### DIFF
--- a/Source/SIMPLib/CoreFilters/ConvertColorToGrayScale.cpp
+++ b/Source/SIMPLib/CoreFilters/ConvertColorToGrayScale.cpp
@@ -44,8 +44,10 @@
 #include <tbb/partitioner.h>
 #include <tbb/task_scheduler_init.h>
 #endif
+#include <algorithm>
 
 #include "SIMPLib/Common/Constants.h"
+#include "SIMPLib/SIMPLibVersion.h"
 #include "SIMPLib/FilterParameters/AbstractFilterParametersReader.h"
 #include "SIMPLib/FilterParameters/DataArraySelectionFilterParameter.h"
 #include "SIMPLib/FilterParameters/IntFilterParameter.h"
@@ -54,14 +56,15 @@
 #include "SIMPLib/FilterParameters/MultiDataArraySelectionFilterParameter.h"
 #include "SIMPLib/FilterParameters/SeparatorFilterParameter.h"
 #include "SIMPLib/FilterParameters/StringFilterParameter.h"
-#include "SIMPLib/SIMPLibVersion.h"
 
-/**
- * @brief The ConvertColorToGrayScaleImpl class implements a threaded algorithm
- * that flattens an RGB array into a grayscale array
- */
 class LuminosityImpl
 {
+private:
+  uint8_t* m_ImageData;
+  uint8_t* m_FlatImageData;
+  FloatVec3_t m_ColorWeights;
+  size_t m_NumComp;
+
 public:
   LuminosityImpl(uint8_t* data, uint8_t* newdata, FloatVec3_t colorWeights, size_t comp)
   : m_ImageData(data)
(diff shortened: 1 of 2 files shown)


Source/SIMPLib/CoreFilters/ConvertColorToGrayScale.h
```diff
@@ -43,17 +42,18 @@
 /**
  * @brief The ConvertColorToGrayScale class. See [Filter documentation](@ref ConvertColorToGrayScale) for details.
  */
-class SIMPLib_EXPORT ConvertColorToGrayScale : public AbstractFilter
+class ConvertColorToGrayScale : public AbstractFilter
```
mmarineBlueQuartz, line 45: SIMPLib_EXPORT is important for loading the class outside of the project it is defined in.  My guess is that there will either be compile errors when building DREAM.3D, or this filter will not be accessible inside of the GUI.

nathanblair: I can't confirm if the filter showed up in the GUI but it built fine. For safety's sake I've re-included that snippet along with making sure to clang-format the files. I'll open up another PR with your included suggestion, thanks Matt!

Source/SIMPLib/CoreFilters/ConvertColorToGrayScale.cpp
```diff
@@ -463,118 +414,31 @@ void ConvertColorToGrayScale::execute()
...
-#ifdef SIMPL_USE_PARALLEL_ALGORITHMS
-    if(doParallel == true)
+    switch (convType)
     {
-        tbb::parallel_for(tbb::blocked_range<size_t>(0, totalPoints), ConvertColorToGrayScaleImpl(m_ImageData, m_FlatImageData, Rfactor, Gfactor, Bfactor, comp), tbb::auto_partitioner());
+      case ConversionType::Luminosity:
+        ParallelWrapper::Run<LuminosityImpl>(LuminosityImpl(inputColorData->getPointer(0), outputGrayData->getPointer(0), m_ColorWeights, comp), totalPoints );
+        break;
+      case ConversionType::Average:
+        ParallelWrapper::Run<LuminosityImpl>(LuminosityImpl(inputColorData->getPointer(0), outputGrayData->getPointer(0), {0.3333f, 0.3333f, 0.3333f}, comp) , totalPoints);
```
mmarineBlueQuartz, line 432: Because of the way floating point numbers are handled, I am still not sold that LuminosityImpl with `{0.3333f, 0.3333f, 0.3333f}` weights is a completely accurate replacement for AverageImpl.  If we are going down this road, I feel using `1 / 3.0f` as a constant would be more accurate than `0.3333f`.  If we were not working with scientific data, I would probably not be as stringent with this.

mmarineBlueQuartz: Since it's merged now, I'd leave it.  This isn't a big enough deal to open another PR about.

nathanblair: Well I'd rather include that SIMPLib_EXPORT statement in another PR anyway. I'm running it through the unit test now for the SIMPLib and pending that succeeding, I'll submit another PR. I don't disagree about the floating point...err...point though. So I've amended that as well and am testing it. The test cases I wrote for the AverageImpl should be robust enough to throw if there were any rounding errors but I think, at least, the point may be more clear using 1/3.

imikejackson:
1/3 is going to come out as .33 to the number of precision as float or double so 0.333333 is good enough for this. At some point the values get flipped back to uint8_t anyways so I think the math should work out to computer precision.

PR #327, opened 2019-02-27
Includes Matthew's suggestions for the ConvertColorToGrayScale filter

nathanblair:
See commit messages.


imikejackson:
Added long ago..

PR #328, opened 2019-03-14
Replacing Cancel boolean with a state enumeration that allows the ability to store Ready, Executing, Canceling, Canceled, and Completed pipeline states

joeykleingers:


mmarineBlueQuartz (changes requested):
Cancel and IsExecuting are not independent values and should not be stored that way.  They are individual states that exist on a spectrum.  Change them to an enum class, and replace the getters and setters to compare states as well as if the state can be set.  A pipeline cannot be cancelled if it isn't being executed, and a pipeline that is still being cancelled or executed probably shouldn't be set to the executing state before returning to an idle state.  From the looks of it, there's no setter for m_Executing, but if we're going to have that state, executing a pipeline should probably check if it can be executed to avoid race conditions.

I'd argue that the FilterPipeline class is a, if not the, core feature that SIMPLib and everything that implements it is built around, so I want to make sure it remains maintainable.

joeykleingers:
Good ideas.  I will work on modifying this and submit another commit.

imikejackson:
I concur that if we need to add state to the Pipeline class it should be done through a class enumeration.

mmarineBlueQuartz (commented):
This looks much better.  There are a few minor things that stood out, though.  What is the purpose of the Completed and Canceled states?  Could we merge Ready, Completed, and Canceled all into one Idle state to avoid bloat and potential bugs when someone only references Ready and not Completed or Canceled?

It would also be nice if we could keep isExecuting and isCanceling (slight name change) that return true if the state matches the corresponding value.  It's like how we have `empty()` so we don't have to write `size() == 0` for readability.  isIdle would probably be a good addition if going down this path.  The easier code is to read, the easier it is to debug when something does go wrong.

joeykleingers:
I added Completed and Canceled states because something outside of FilterPipeline may want to know whether the most recent execution was canceled or not.  Once the pipeline goes from Executing to Idle, there isn't a way to interrogate how the most recent execution ended, unless we store that result somewhere and provide a helper function that can be used to return that result.

joeykleingers:
We could add an ExecutionResult class enumeration that simply has values "Not_Executed" (need better name), "Canceled", and "Completed" and allow outside classes to ask the FilterPipeline for its ExecutionResult value.

mmarineBlueQuartz:
We have signals to notify anything listening whether a pipeline was completed or canceled.  Otherwise, those values should be put in their own enum as the result of a pipeline execution is not a state of execution.  If we do save that result, it needs to be its own value to prevent state from losing its meaning.  Then add an InProgress or Invalid value to the enum for when the pipeline has not yet been run or is in the process of being executed.

joeykleingers:
Is there ever a case where an object that's created after a pipeline executes (and therefore doesn't get to hook up to the pipelineCompleted and pipelineCanceled signals) will want to know the result of the pipeline execution?  In other words, is there a benefit to being able to get that information after a pipeline has executed instead of having to be hooked up to the signal when the pipeline is finishing?

mmarineBlueQuartz:
Unit tests we haven't written could use it, but even in cases where it is connected via signals and slots, having an enum to consistently store the value could be useful.  For instance, in part of my multi-pipeline visualization branch, I access errors and warnings through the pipeline model, and having a consistent enum to return without reinventing the wheel between GUI/non-GUI implementations would be optimal, even if the value isn't saved in the pipeline.  Having that type at the same level as where the value would be made available helps prevent issues where we redefine the same functionality in different places (especially bad if they behave differently), and just keeps the code cleaner.

mmarineBlueQuartz (approved):
This looks good.  Ideally, there would be switch statements for enum checks, but I'm happy with this.

PR #330, opened 2019-03-19
Fix crash in RawBinaryReader when using large SkipHeaderBytes values

### LINKED ISSUE #329: RawBinaryReader Crashes when large values to skip are used.
Using a large "Skip Header Bytes" value causes a crash.

imikejackson:
The issue appeared because the assumption was that the amount of data eventually
read would be larger than the number of header bytes to skip. Implemented a UInt64
filter parameter and matching widget. This new filter parameter is used for the
RawBinaryReader so that the use can have offsets larger than 2GB (2^31). The original
value was stored in an int32_t which would eventually be inadequate.

#fixes #329. closes #329. updates #329

### DIFF
--- a/Source/SIMPLib/CoreFilters/RawBinaryReader.cpp
+++ b/Source/SIMPLib/CoreFilters/RawBinaryReader.cpp
@@ -99,7 +99,7 @@ int32_t SanityCheckFileSizeVersusAllocatedSize(size_t allocatedBytes, size_t fil
 template <typename T> int32_t readBinaryFile(RawBinaryReader* filter)
 {
 
-  DataArray<T>::Pointer p = filter->getDataContainerArray()->getPrereqIDataArrayFromPath<DataArray<T>, AbstractFilter>(filter, filter->getCreatedAttributeArrayPath());
+  typename DataArray<T>::Pointer p = filter->getDataContainerArray()->getPrereqIDataArrayFromPath<DataArray<T>, AbstractFilter>(filter, filter->getCreatedAttributeArrayPath());
   QString filename = filter->getInputFile();
   int64_t skipHeaderBytes = filter->getSkipHeaderBytes();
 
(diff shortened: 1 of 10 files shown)


joeykleingers (changes requested):
Generally looks good.  The only thing I ask is that the 2019 BlueQuartz Software license be added to every header and implementation file in this commit.  It looks like some of the files do have it and some are using the older license.

mmarineBlueQuartz (approved):
This looks good.  I had some problems with RawBinaryReader using size_t for skipHeaderBytes in SanityCheckFileSizeVersusAllocatedSize and readBinaryFile instead of uint64_t like the corresponding filter parameter, but since reading the actual methods and finding the only real usage of those parameters cast to size_t anyways, I am okay with this.

imikejackson:
@mmarineBlueQuartz What problems were you seeing? This code might not actually work on Linux and/or Windows. I need to test there...

mmarineBlueQuartz:
It's not exactly a compile issue, but size_t is not a hard data size.  It can mean different things under different compile conditions, moreso than `int`.  I'm most familiar with it being defined differently between linux and otherwise (unsigned long vs unsigned long long, but I think that's a difference in long long definition), but when I looked into it more to be sure about mixing uint64_t and size_t, it compiles down to an unsigned integer of the same bit-size as being compiled for (32 / 64-bit).  Since we aren't compiling for 32-bit, and I'm not expecting 128-bit for another decade or more, this conversion should be fine, albeit a bit uncomfortable.  However, since the only real use of the parameter would have been casting to a size_t, there's no point in delaying the cast until halfway through the method.

imikejackson:
The requested changes have been made in addition to a few other rewrites. Let me know your thoughts on these changes

imikejackson:
Tested on Windows and passes Unit Tests.

mmarineBlueQuartz (approved):
I like these new changes, especially the switch statement.  That made it a lot easier to read.

imikejackson:
PR Comment.

joeykleingers (approved):
Looks great!  I'm starting to come around to how switch statements look.  Before, they always looked super odd to me, but I can see how they can be a lot more readable.  I think it's partly that I cringe at using the 'break' keyword anywhere in my code.

imikejackson:
The switches are really useful with the C++11 class enums as the compiler will tell you if you missed one of the enums.

imikejackson:
This PR has now been tested on each of the platforms.

PR #331, opened 2019-03-25
WIP: Addressed arithmetic overflow and other code analysis warnings

mmarineBlueQuartz:
* Initialized PipelineMessage index in constructors.
* Marked a possible coding error in ImportAsciiDataArray where two identical error conditions are checked using || operator.
* Fixed an incorrect cast in ReadASCIIData where a double is cast to a float for a method parameter that is expecting a double.  Due to multiple calculations of the same value, it was saved to a constant before the first calculation was required.
* Changed a for loop variable in geometry compute methods to avoid default castings when dereferencing a pointer multiple times per iteration.
* Changed geometry requirements in addAttributeMatrix methods from using || for enum checks to &&.  The previous versions could never be false and prevented the entire method call from doing anything.
* Fixed a possible overflow when calculating ImageGeom and RectGridGeom coords.  The previous versions would have overflowed before casting to a larger size.
* Updated variables used to store min and max float values to constexpr to take advantage of the fact that they are known at compile time and are never modified.
* Removed unnecessary casts in SIMPLibRandom where doubles were cast to float and then saved as a double.
* Fixed ColorTable::GetColorTable due to the current implementation going out of array bounds.  The 2D array float color[8][3] was being accessed with indices [8][2].


Source/SIMPLib/CoreFilters/ImportAsciDataArray.cpp
```diff
@@ -190,7 +190,7 @@ template <typename T, typename K> int32_t readAsciFile(typename DataArray<T>::Po
       {
         return err;
       }
-      if(err == RBR_READ_ERROR || err == RBR_READ_ERROR)
+      if(err == RBR_READ_ERROR || err == RBR_READ_ERROR) // Identical
```
imikejackson: They are the same but I wonder if the code should be checking another error condition?

Source/SIMPLib/CoreFilters/ImportAsciDataArray.cpp
```diff
@@ -208,7 +208,7 @@ template <typename T, typename K> int32_t readAsciFile(typename DataArray<T>::Po
...
```
imikejackson: Same for this one...

Source/SIMPLib/Utilities/ColorTable.cpp
```diff
@@ -68,21 +69,32 @@ void SIMPLColorTable::GetColorTable(int numColors, QVector<float>& colors)
       {255.0f / 255.0f, 0.0f / 255.0f, 0.0f / 255.0f} // red
   };
 
-  float val = 0.0f, r = 0.0f, g = 0.0f, b = 0.0f;
   float step = 1.0 / float(numColors);
   float nodeStep = 1.0f / float(numColorNodes - 1);
-  for(int i = 0; i < (numColors); i++)
+  for(int i = 0; i < numColors; i++)
```
imikejackson, line 75: Can you use new style C++ casts rather than old "C" style casts?

imikejackson:
Generally looks good. Wish I had a few of these before the release on Friday. We will get them on next months release.

imikejackson:
Can you rebase/merge the latest develop in with this branch and update the PR?

PR #332, opened 2019-03-26
Fixed SVPipelineView execution lock on error

mmarineBlueQuartz:
* Fixed a bug where, if a FilterPipeline failed during execution, the pipeline view would be permanently locked in the Executing and Canceling states.  During this, filters cannot be added or reordered, and all filter input widgets are disabled.  The user can get to the Canceling state, but once there, cannot leave.
* Added Failed value to FilterPipeline::ExecutionResult enum.
* SVPipelineView now uses a switch statement instead of if else blocks to handle the pipeline's execution result.

Updates BlueQuartzSoftware/DREAM3D#893
Fixes BlueQuartzSoftware/DREAM3D#893


imikejackson:
Looks good. Merge it in..

PR #333, opened 2019-03-28
Feature/MessagesRedesign

joeykleingers:
* Replacing PipelineMessage API with an AbstractMessage superclass and subclasses for every type of message.  This eliminates the need for the MessageType enumeration.

* Implementing a new, cleaner and more reusable API that observers can implement to be able to process specific types of messages.  Take a look at the IssuesWidget and IssuesWidgetMessageHandler for a good example of this.  Observers can decide which types of messages they are interested in by subclassing the AbstractMessageHandler class and reimplement the appropriate processMessage() methods.

* Passing AbstractMessages from observable object to observer object using a shared pointer instead of by value. This allows observers to store the messages with all the relevant subclass data if they prefer.

* Eliminating the use of humanLabel in Observable's and AbstractFilter's notify****Message method parameters.  In AbstractFilter, these methods already set the human label into the message.

* It is no longer necessary to use setErrorCondition() or setWarningCondition() outside of the notify****Message methods.  The notify methods will automatically set the condition to the code that is passed in.

* Filters that emit progress messages while they are in an executing pipeline have their progress values automatically converted into overall pipeline progress values and re-emitted as pipeline progress messages.  This allows filters to update the overall progress of the pipeline while they are running.  Check out FilterPipelineMessageHandler at the top of the FilterPipeline implementation file.


Source/SIMPLib/Common/IObserver.cpp
```diff
@@ -50,34 +50,10 @@ IObserver::~IObserver() = default;
 // -----------------------------------------------------------------------------
 //
 // -----------------------------------------------------------------------------
-void IObserver::processPipelineMessage(const PipelineMessage& pm)
+void IObserver::processPipelineMessage(AbstractMessage::Pointer pm)
```
mmarineBlueQuartz: This should still be a const reference.

Source/SIMPLib/CoreFilters/MoveMultiData.cpp
```diff
@@ -195,19 +194,17 @@ void MoveMultiData::dataCheck()
...
         return;
       }
       // Source and Destination match
       if(daSrcPaths[i].hasSameAttributeMatrix(amDestPath))
       {
-        setWarningCondition(-11020);
-        QString ss = QObject::tr("The source and destination Attribute Matrix are the same.  Is this what you meant to do?");
-        notifyWarningMessage(getHumanLabel(), ss, getWarningCondition());
+                QString ss = QObject::tr("The source and destination Attribute Matrix are the same.  Is this what you meant to do?");
+        notifyWarningMessage("", ss, -11020);
```
mmarineBlueQuartz: Check MoveData.cpp and MoveMultiData.cpp.  I think you are using tabs, or your formatting is just messed up.

joeykleingers: Yeah I did a grep for every place that uses setErrorCondition/notifyErrorMessage and setWarningCondition/notifyWarningMessage and did a mass replacement to remove the set******Condition calls and put whatever code was passed to that call as the code passed to notify******Message.  So there may have been a few that got a few lines of formatting messed up.

joeykleingers: After I did the grep, I made sure to search the whole code base for mentions of those in case I missed a few.

joeykleingers: There's a few places where we set the condition but never emit a message, so I wasn't sure what to do with those, so I left them be.

Source/SIMPLib/Filtering/FilterPipeline.cpp
```diff
@@ -872,34 +867,28 @@ DataContainerArray::Pointer FilterPipeline::execute()
...
   else
   {
     // This should never get here
-    completeMessage.setText("Unsupported Pipeline Execution Result");
-    emit pipelineGeneratedMessage(completeMessage);
-
-    setErrorCondition(-210);
-    completeMessage.setType(PipelineMessage::MessageType::Error);
-    completeMessage.setCode(getErrorCondition());
+    notifyErrorMessage("", "Unsupported Pipeline Execution Result", -210);
```
mmarineBlueQuartz: Enum checks should use switch statements.  They're generally more readable and shorter.  And if you end up where you shouldn't be, a good switch statement makes it very obvious where that is, unlike `else`.

Also, remove the `// This should never get here` comment.  We don't handle FilterPipeline error codes, so this is going to silently fail or worse.  If we absolutely shouldn't be there, throw an exception or in some way, shape, or form, make it obvious that we've gone off the rails.  The last time I hit a comment like that (and it was in this class), it was tracking down a bug where the GUI was being permanently locked but nothing was causing a message to be displayed or a crash where the debugger could say where we went off into undefined territory.  If you shouldn't be here, make it obvious and very loud when you do.

Source/SIMPLib/Messages/AbstractErrorMessage.h
```diff
@@ -0,0 +1,84 @@
+/* ============================================================================
+ * Copyright (c) 2009-2016 BlueQuartz Software, LLC
```
mmarineBlueQuartz: 2019-2019

Source/SIMPLib/Messages/AbstractErrorMessage.h
```diff
@@ -0,0 +1,84 @@
...
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+ * USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * The code contained herein was partially funded by the followig contracts:
+ *    United States Air Force Prime Contract FA8650-07-D-5800
+ *    United States Air Force Prime Contract FA8650-10-D-5210
+ *    United States Prime Contract Navy N00173-07-C-2068
```
mmarineBlueQuartz, line 32: As I understand it, new files aren't under these prime contracts.  You might want to ask Mike what this is under.

Source/SIMPLib/Messages/AbstractMessageHandler.h
```diff
@@ -0,0 +1,80 @@
...
+
+    virtual void processMessage(PipelineErrorMessage* msg) const;
+    virtual void processMessage(PipelineProgressMessage* msg) const;
+    virtual void processMessage(PipelineStatusMessage* msg) const;
+    virtual void processMessage(PipelineWarningMessage* msg) const;
+
+    virtual void processMessage(FilterErrorMessage* msg) const;
+    virtual void processMessage(FilterProgressMessage* msg) const;
+    virtual void processMessage(FilterStatusMessage* msg) const;
+    virtual void processMessage(FilterWarningMessage* msg) const;
```
mmarineBlueQuartz: Is there isn't a case where we should be editing the messages in message handler?  If not, these should be const parameters.  My understanding so far is that we should not be changing the message after it has been created.

Source/SIMPLib/Messages/FilterErrorMessage.cpp
```diff
@@ -0,0 +1,83 @@
...
+*    United States Air Force Prime Contract FA8650-07-D-5800
+*    United States Air Force Prime Contract FA8650-10-D-5210
+*    United States Prime Contract Navy N00173-07-C-2068
+*
+* ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ */
+
+#include "FilterErrorMessage.h"
+
+#include <QtCore/QMetaType>
+#include <QtCore/QString>
```
mmarineBlueQuartz: Why do we have QString included in our .cpp when it's used in the .h?  This is already included thanks to AbstractMessage.h

Source/SIMPLib/Messages/AbstractErrorMessage.h
imikejackson: If the time is being billed to a prime contract then the code is under a  prime contract. This is one of those cases.

Source/SIMPLib/Messages/GenericErrorMessage.h
```diff
@@ -0,0 +1,79 @@
...
+*    United States Air Force Prime Contract FA8650-10-D-5210
+*    United States Prime Contract Navy N00173-07-C-2068
+*
+* ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ */
+
+#pragma once
+
+#include "SIMPLib/Messages/VisitableErrorMessage.h"
+
+class AbstractMessageHandler;
```
mmarineBlueQuartz: You don't need AbstractMessageHandler declared here.  It isn't used in the header, and it is included in the source file.  If something includes GenericErrorMessage requires AbstractMessageHandler to be defined, it should be defined where it is required rather than everywhere it's not.

Source/SIMPLib/Messages/GenericStatusMessage.cpp
```diff
@@ -0,0 +1,86 @@
...
+*    United States Prime Contract Navy N00173-07-C-2068
+*
+* ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ */
+
+#include "GenericStatusMessage.h"
+
+#include <QtCore/QMetaType>
+#include <QtCore/QString>
+
+#include "SIMPLib/Messages/AbstractMessageHandler.h"
```
mmarineBlueQuartz: You don't need to include AbstractMessageHandler.

Source/SIMPLib/Plugin/SIMPLPluginConstants.h
```diff
@@ -21,7 +21,7 @@ const QString StatusLog("StatusLog");
 const QString OutputLinks("OutputLinks");
 const QString Message("Message");
 const QString Code("Code");
-const QString FilterHumanLabel("FilterHumanLabel");
+const QString HumanLabel("HumanLabel");
```
mmarineBlueQuartz: I've got this strange feeling that this needlessly breaks everyone's pipelines.

imikejackson: I'm not sure if it will or not. We cue off of the class name or the UUID when reading pipelines. The Human Label is written into the pipeline to help those that want to edit the pipeline but I don't think it will break anything per se. I am wondering why we needed to change that constant?

mmarineBlueQuartz: I guess you're right.  The pipeline json has underscores in it.  I missed that part.

Source/SIMPLib/Messages/AbstractMessageHandler.h
imikejackson: From the name of the method is it "ProcessMessage" with no indication that any editing of the object would occur so make the arguments into const FilterErrorMessage* msg. Also add DOxygen documentation for these methods to amply describe what they are doing.

Source/SIMPLib/Messages/AbstractErrorMessage.h
imikejackson: 2009-1019 since the file was first created back in 2009

mmarineBlueQuartz: Can we go back to the future real quick.  This isn't the direction I wanted my life to go.

Source/SIMPLib/Messages/AbstractErrorMessage.h
mmarineBlueQuartz: I agree with the billing and code being under said contract, but are we under one of these three specifically?  I was under the impression it changed, or at least not all three of these.

mmarineBlueQuartz (changes requested):
- You specify that you implemented double dispatch, but that isn't accurate from what I can tell.  You only implemented single dispatch.  Update the PR message because I know how easily this can propagate into the commit message and eventual technical report.
- You should probably remove your abstract `Visitable*Message` templated classes.  From what I can tell, there is nothing notable added by these classes other than confusion about what messages are what.  The necessary Visit method is already declared in AbstractMessage, meaning every message is visitable.  Furthermore, every `Visitable*Message` is an `Abstract*Message`, but no `Abstract*Message` is not a `Visitable*Message`.  Simply add your definition to the visit method to each concrete implementation of AbstractMessage, and we won't have to worry about if something is a `Visitable*Message` or `Abstract*Message` because those are synonymous from what I can tell.
- (In the future, not necessarily this PR) Avoid including or declaring classes you are not using.  That happened a lot as I read through the new files.  Our drives at work are NVMe interfaced because it's so much faster to access all of our included files while compiling.  Adding unnecessary includes makes that process slower for everyone.  C++ is built around the premise of pay for what you need rather than pay for everything.  If you don't need a definition, don't include one.  If you don't need a declaration, don't provide one.  Pay attention to what you need and leave it at that.

Edited: Fixed random italics while preserving (re-introducing?) a templated naming convention.

Source/SIMPLib/Plugin/SIMPLPluginConstants.h
joeykleingers: I'm going to change it back just in case.

joeykleingers:
> * You specify that you implemented double dispatch, but that isn't accurate from what I can tell.  You only implemented single dispatch.  Update the PR message because I know how easily this can propagate into the commit message and eventual technical report.

Is calling an overloaded visit method on an AbstractMessage, and then calling an overloaded processMessage function on the AbstractMessageHandler instance that you passed in not double dispatch?

joeykleingers:
> * You should probably remove your abstract Visitable_Message templated classes.  From what I can tell, there is nothing notable added by these classes other than confusion about what messages are what.  The necessary Visit method is already declared in AbstractMessage, meaning every message is visitable.  Furthermore, every Visitable_Message is an Abstract_Message, but no Abstract_Message is not a Visitable_Message.  Simply add your definition to the visit method to each concrete implementation of AbstractMessage, and we won't have to worry about if something is a Visitable_Message or Abstract*Message because those are synonymous from what I can tell.

I did this to make it easier in the future for people to add other types of messages.  Without it, each time you add a message class, you have to do a copy-paste of the visit method to reimplement it in your class.  Whereas having it in a templated superclass like that saves you that step.  I agree that the templated class can probably be renamed to something better, but I was just trying to make it as "idiot proof" as possible to add a new message class.

joeykleingers:
> * (In the future, not necessarily this PR) Avoid including or declaring classes you are not using.  That happened a lot as I read through the new files.  Our drives at work are NVMe interfaced because it's so much faster to access all of our included files while compiling.  Adding unnecessary includes makes that process slower for everyone.  C++ is built around the premise of pay for what you need rather than pay for everything.  If you don't need a definition, don't include one.  If you don't need a declaration, don't provide one.  Pay attention to what you need and leave it at that.

I agree 100%.  A lot of it was stuff that was leftover from implementations that I was trying out and ended up changing.  I will make sure to look at each file to make sure I don't have any leftovers next time.

Source/SIMPLib/Messages/AbstractErrorMessage.h
imikejackson: Parts of the code have been written under all three of those prime contracts. For every prime contract that the code is written under it should be listed.

mmarineBlueQuartz:
> I did this to make it easier in the future for people to add other types of messages. Without it, each time you add a message class, you have to do a copy-paste of the visit method to reimplement it in your class. Whereas having it in a templated superclass like that saves you that step. I agree that the templated class can probably be renamed to something better, but I was just trying to make it as "idiot proof" as possible to add a new message class.

That made it less idiot proof because now it's even harder to ~~easily~~ understand how things work or what type they should be using.  Copy-pasting a literal 1-liner is not worth that.  Don't bother renaming the templated class.  REMOVE it.

Update: I had to go through the code to determine if it was even legal because the inheritance did not make sense at first.  If you have to **derive from** a class that is templated **on the derived class**, something has probably gone wrong.

Source/SIMPLib/Common/Observable.h
```diff
@@ -55,7 +55,10 @@ class SIMPLib_EXPORT Observable : public QObject
 {
     Q_OBJECT
     PYB11_CREATE_BINDINGS(Observable)
-    PYB11_METHOD(void notifyErrorMessage ARGS humanLabel ss code)
+    PYB11_METHOD(void notifyErrorMessage ARGS humanLabel msg code)
+    PYB11_METHOD(void notifyWarningMessage ARGS humanLabel msg code)
+    PYB11_METHOD(void notifyStatusMessage ARGS humanLabel msg)
+    PYB11_METHOD(void notifyProgressMessage ARGS humanLabel msg progress)
```
mmarineBlueQuartz: These methods no longer exist.  You might also want to consider wrapping the WithPrefix methods.

imikejackson: Please wrap the appropriate methods with the PYB11_* Macros if possible.

Source/SIMPLib/CoreFilters/MultiThresholdObjects.cpp
```diff
@@ -213,13 +209,12 @@ void MultiThresholdObjects::execute()
   {
     ThresholdFilterHelper filter(static_cast<SIMPL::Comparison::Enumeration>(comp_0.compOperator), comp_0.compValue, m_DestinationPtr.lock().get());
     // Run the first threshold and store the results in our output array
-    err = filter.execute(m->getAttributeMatrix(amName)->getAttributeArray(comp_0.attributeArrayName).get(), m_DestinationPtr.lock().get());
+    int err = filter.execute(m->getAttributeMatrix(amName)->getAttributeArray(comp_0.attributeArrayName).get(), m_DestinationPtr.lock().get());
```
mmarineBlueQuartz: This isn't the first occurrence, but you probably shouldn't replace `int32_t` with `int`, even if they technically mean the same thing.  int32_t declares a stronger intent and the type is intentionally 32-bits instead of the default int type.  There are plenty of places where we use the default int type because of Qt where int does not make any sense (i.e. this array size is -5), so declaring this as int32_t tells the reader that this is intentional and it actually should be the type we say it is.  I'm not so worried about the places where it wasn't int32_t, but erasing the intent should be avoided where possible.

Source/SIMPLib/CoreFilters/RenameAttributeArray.cpp
```diff
@@ -118,44 +117,41 @@ void RenameAttributeArray::dataCheck()
...
+    }
+    case SUCCESS:
+    {
+      clearErrorCondition();
+      clearWarningCondition();
+      DataArrayPath newPath = getSelectedArrayPath();
+      newPath.setDataArrayName(m_NewArrayName);
+      addPathRename(getSelectedArrayPath(), newPath);
+      return;
+    }
```
mmarineBlueQuartz: Something happened to your formatting here.  None of these cases should be indented, at least not compared to the preceding switch statement.

Source/SIMPLib/Filtering/AbstractFilter.h
```diff
@@ -87,9 +87,8 @@ class SIMPLib_EXPORT AbstractFilter : public Observable
   PYB11_PROPERTY(QString CompiledLibraryName READ getCompiledLibraryName)
   PYB11_PROPERTY(bool Cancel READ getCancel WRITE setCancel)
   PYB11_PROPERTY(bool Enabled READ getEnabled WRITE setEnabled)
-  PYB11_PROPERTY(QString MessagePrefix READ getMessagePrefix WRITE setMessagePrefix)
-  PYB11_PROPERTY(int ErrorCondition READ getErrorCondition WRITE setErrorCondition)
-  PYB11_PROPERTY(int WarningCondition READ getWarningCondition WRITE setWarningCondition)
+  PYB11_PROPERTY(int ErrorCondition READ getErrorCondition)
+  PYB11_PROPERTY(int WarningCondition READ getWarningCondition)
```
mmarineBlueQuartz: getErrorCondition and getWarningCondition no longer exist

Source/SIMPLib/Filtering/FilterPipeline.cpp
```diff
@@ -45,12 +54,43 @@
...
+      float filterProgressStep = (1.0f / m_Pipeline->size()) * filterProgress / 100.0f;
+      int pipelineProgress = static_cast<int>((static_cast<float>(msg->getPipelineIndex()) / (m_Pipeline->size()) + filterProgressStep) * 100.0f);
+      m_Pipeline->notifyProgressMessage(pipelineProgress, "");
+    }
+
+  private:
+    FilterPipeline* m_Pipeline = nullptr;
+};
+
+struct PipelineIdleException : public std::exception
```
mmarineBlueQuartz: Syntactically correct, but we're deriving a non-POD struct from a class?

Source/SIMPLib/Filtering/FilterPipeline.cpp
```diff
@@ -45,12 +54,43 @@
...
+{
+   const char* what () const throw ()
```
mmarineBlueQuartz: The `const throw()` was removed in C++11.  Use `const noexcept`

Source/SIMPLib/Filtering/FilterPipeline.cpp
```diff
@@ -273,8 +313,8 @@ void FilterPipeline::fromJson(const QJsonObject& json, IObserver* obs)
                                  Possible reasons include a name change of the filter, plugin not loading or a simple spelling mistake? A \
                                                                                                                                         blank filter has been inserted in its place. Possible error message is: %1")
                          .arg(filterName);
-        PipelineMessage pm(filterName, ss, -66066, PipelineMessage::MessageType::Error);
-        pm.setPrefix("JsonFilterParametersReader::ReadPipelineFromFile()");
+        QString prefix = "JsonFilterParametersReader::ReadPipelineFromFile()";
+        PipelineErrorMessage::Pointer pm = PipelineErrorMessage::New(getName(), ss, -66066);
         obs->processPipelineMessage(pm);
```
mmarineBlueQuartz, line 318: The QString prefix variable is never used.

mmarineBlueQuartz: Also, how did that become a variable?

joeykleingers: It's a prefix that was leftover after removing the prefix as a message variable.  There are one or two places where we create the message ourselves and pass it into the processPipelineMessage method of an observer that we have access to.  Should probably change that to just use the setErrorCondition method and hook it up so that the observer's slot automatically gets called.

Source/SIMPLib/Filtering/FilterPipeline.cpp
```diff
@@ -869,34 +891,33 @@ DataContainerArray::Pointer FilterPipeline::execute(DataContainerArray::Pointer
...
   disconnectSignalsSlots();
 
-  PipelineMessage completeMessage("", "", 0, PipelineMessage::MessageType::StatusMessage, -1);
-  if(m_State == FilterPipeline::State::Canceling)
-  {
-    completeMessage.setText("Pipeline Canceled");
-    m_ExecutionResult = FilterPipeline::ExecutionResult::Canceled;
-  }
-  else if(m_State == FilterPipeline::State::Executing)
+  switch(m_State)
```
mmarineBlueQuartz, line 900: Check your switch state formatting.  Cases should be at the same indentation as the switch.

mmarineBlueQuartz: You also don't need curly brackets inside case blocks unless you create a variable inside it.

Source/SIMPLib/Geometry/IGeometry.h
```diff
@@ -76,7 +76,6 @@ class SIMPLib_EXPORT IGeometry : public Observable
   PYB11_METHOD(Type getGeometryType)
   PYB11_METHOD(QString getGeometryTypeAsString)
   PYB11_METHOD(QString getInfoString ARGS formatType)
-  PYB11_PROPERTY(QString MessagePrefix READ getMessagePrefix WRITE setMessagePrefix)
   PYB11_PROPERTY(QString MessageTitle READ getMessageTitle WRITE setMessageTitle)
   PYB11_PROPERTY(QString MessageLabel READ getMessageLabel WRITE setMessageLabel)
```
mmarineBlueQuartz: MessageTitle and MessageLabel were removed as well.

Source/SIMPLib/REST/PipelineListener.h
```diff
@@ -18,28 +25,33 @@ class SIMPLib_EXPORT PipelineListener : public QObject, public IObserver
...
-  QString getStandardOutputLog();
 
 public slots:
-  void processPipelineMessage(const PipelineMessage& pm);
+  void processPipelineMessage(const AbstractMessage::Pointer& pm);
 
 private:
-  std::vector<PipelineMessage> m_Messages;
+  std::vector<AbstractMessage::Pointer> m_SharedMessages;
+  std::vector<const AbstractMessage*> m_Messages;
```
mmarineBlueQuartz: Every other variable holding a collection of messages has something describing why the collection matters and what it is for.  In that context, m_Messages sounds like a variable that someone forgot to delete.  Consider renaming to m_AllMessages, or if that isn't accurate, some way of describing why it exists.  Consider updating the corresponding method to match.  Simply leaving it at Messages is too generic with respect to everything else and can easily be misused.   For instance, someone can parse through getMessages() and look for ErrorMessages because if someone is looking for messages, there is nothing in the name signifying there might be other collections already sorted.

Source/SIMPLib/REST/PipelineListenerMessageHandler.h
```diff
@@ -0,0 +1,68 @@
+#pragma once
```
mmarineBlueQuartz, line 37: The copyright header is missing

Source/SIMPLib/REST/PipelineListenerMessageHandler.h
```diff
@@ -0,0 +1,68 @@
...
+class SIMPLib_EXPORT PipelineListenerMessageHandler : public AbstractMessageHandler
+{
+  public:
+    PipelineListenerMessageHandler(PipelineListener* listener);
+
+    /**
+     * @brief Streams incoming PipelineErrorMessages to the error log and stores the message
+     * for later use.
+     */
+    virtual void processMessage(const PipelineErrorMessage* msg) const override;
```
mmarineBlueQuartz: Clang-format this file

mmarineBlueQuartz (changes requested):
You are inconsistent in your use of `const QString &name` and `const QString& name`, often using both on the same line.  Pick one and only one.  Our clang-format rules will set it to `const QString& name` if you are not sure which you want.  I also only noticed partway through, but it seems none of the new header files / class definitions you created are compliant with the specified clang-formatting rules.  You might want to update your IDE defaults or clang-format on commit.

I mentioned it in one of the comments I left, but getMessages() is too generic when there are several other getMessages methods for specific purposes.  Update the name of getMessages() to describe how it relates to the other collections.  I noticed that naming issue in at least two different classes when I came back to it this morning.

mmarineBlueQuartz (approved):
This looks good to me.

PR #337, opened 2019-03-30
Fix loading of plugins for PipelineRunner on Linux

imikejackson:
Ensure that the lib directory is searched for plugins

### DIFF
--- a/Source/SIMPLib/Plugin/SIMPLibPluginLoader.cpp
+++ b/Source/SIMPLib/Plugin/SIMPLibPluginLoader.cpp
@@ -154,30 +154,30 @@ void SIMPLibPluginLoader::LoadPluginFilters(FilterManager* filterManager, bool q
     {
       thePath = aPluginDir.absolutePath();
       pluginDirs << thePath;
-      aPluginDir.cdUp(); // Move back up a directory level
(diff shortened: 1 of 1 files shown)


Source/SIMPLib/Plugin/SIMPLibPluginLoader.cpp
```diff
@@ -151,6 +160,25 @@ void SIMPLibPluginLoader::LoadPluginFilters(FilterManager* filterManager, bool q
...
+    {
+      thePath = aPluginDir.absolutePath();
+      pluginDirs << thePath;
+      aPluginDir.cdUp(); // Move back up a directory level
+      int no_error = chdir(aPluginDir.absolutePath().toLatin1().constData());
+      if(no_error < 0)
+      {
+        if(!quiet) qDebug() << "Could not set the working directory.";
+      }
+      if(!quiet)
```
mmarineBlueQuartz, line 172: I am not entirely certain, but this looks like it should be an `else if` considering the preceding error check.

imikejackson: Adjusted order to make it clearer when error/status messages are printed.

mmarineBlueQuartz (approved):
This looks good to me.

PR #338, opened 2019-04-11
Reduce the creation of filters through factories when not needed.

imikejackson:
When plugins are loaded, a FilterFactory<T> is created and stored for each filter that is in the plugin. If the developer only needs items such as human label, Html summary or other textual items then it is better and quicker to just use the FilterFactory itself instead of instantiating an entirely new Filter object. This leads to a speed up when launching currently.


mmarineBlueQuartz (approved):
This looks good to me.

PR #339, opened 2019-04-15
Feature/linked path parameter

mmarineBlueQuartz:
* Added LinkedPathCreationFilterParameter for creating AttributeMatrix and DataArray.  This filter parameter sets a QString property for the target name.
* Updated filters to use LinkedPathCreationFilterParameter.
* Creating an AttributeMatrix with linked DataContainer requires one additional value for getting the DataContainer property.  This property can be a DataArrayPath or QString.
* Creating a DataArray with linked AttributeMatrix requires two additional value for getting the DataContainer and AttributeMatrix properties.  These properties can be DataArrayPath or QString. DataArrayPath properties will only access the name corresponding to the target level.  For example, two DataArrayPath properties will grab the DataContainer name from the first and the AttributeMatrix name from the second.
* Creating a DataArray from an ComparisonInputsAdvanced property extracts the DataContainer and AttributeMatrix from that linked property.  This is only used in MultiThresholdObjects2.


imikejackson (commented):
+ Need to update DevHelper templates to include this new Filter Parameter.
+ New files should only have a copy right date of 2019-2019
+ The  prime contract should United States Air Force Prime Contract FA8650-15-D-5280
+ DataPathLabel.cpp GetInactiveColor() has a hard coded color which may or may not be a good idea based on all the different themes that we have going now.

imikejackson:
Screen Shot 2019-04-18 at 10 44 31 AM

I think there needs to be some styling on the UI widget. With it just being plain it is hard to tell where the Human label ends and the Data Array path begins. I have no ideas what would be good at this point. Or does my style sheet need to be updated?

mmarineBlueQuartz:
There is styling, but I hadn't push the feature/LinkedPathParameter branch on BrandedDREAM3D.  each part of the path is colored by the corresponding DataArrayPath color for the theme.

imikejackson:
Let's push that branch. This also serves to see what it looks like "unstyled" which is what it would look like by default. I think just some spacing on the widgets might fix it.

imikejackson:
Screen Shot 2019-04-18 at 11 15 37 AM
 Grabbed the style from Midnight.css for that screen grab. See if you can do the same for the Default.css inside of SIMPL/Resources....  and see if that looks acceptable on Windows.

### Other Comments ###

Screen Shot 2019-04-18 at 11 20 31 AM

Looking at the Midnight Theme I think having colored text in the text box makes the text box look too much like the DataArraySelectionWidget (and other widgets of that vein) which might cause confusion to the user, i.e., what is a button and what is a text field. Load up the number 8 pipeline under the Reconstruction prebuilt and select the "Merge Twins" filter. This is a good example of what I am talking about. We could try setting a background color for the text area instead (just a really light shade of the blue/purple/green) and see how that looks. I think the UX folks would tell us to go gentle on the use of color.. 

### Dark Theme ###
Screen Shot 2019-04-18 at 11 23 32 AM
The "Dark" theme works better because the counter-color is orange instead of blue of the "Midnight"theme. With the "Dark" theme I like this styling better because it now becomes clear what is going on with the colors and the colors are not ambiguous.

mmarineBlueQuartz:
The DataArrayPathSelectionWidget does not color its text, but I found this best describes what the text input is rather than leave it to the default color (Midnight makes all of them look like DataArray names).  This caused the most confusion in my own usage.  If we don't color the text, we have to avoid allowing the data structure colors matching the theme at all, or remove data structure colors entirely.

If we are having difficulties differentiating buttons and text fields, that should be affecting the normal line edits far more as there is more similarity there.

imikejackson:
Also getting this warning when running:

QCssParser::parseHexColor: Unknown color name '#404040_disabled'

*fixed* in the PR against your fork @mmarineBlueQuartz

imikejackson:
Screen Shot 2019-04-18 at 1 09 53 PM

Added a slightly lighter background to the DataArrayPathSelectioWidget (using the theme engine). I think this helps differentiate between a button and a text field.

imikejackson:
I think we still need to apply the text colors to the "Default" style.

imikejackson:
Do all the unit tests pass?

mmarineBlueQuartz:
The unit tests pass.

imikejackson:
Merge it

PR #340, opened 2019-04-17
Add DataStructure context menu

mmarineBlueQuartz:
* Expand All
* Collapse All
* Collapse All but target index, children, and parents
   * Target requires children


imikejackson (commented):
Generally looks good and the functionality is there. I was thinking of having some sort of button at the top so it is more obvious to the user that this operation is available. I will agree up front that having a toolbar (small one) with a single button on it might look a bit strange right now but in the future I bet we could find other items to put in there.
The "tool bar" should probably be pretty small, maybe only 16~20 pixels high. We can come up with an image for the button. Take a a look at VS Code's icon to get an idea of what it should look like.

imikejackson:
Merged in through another commit

imikejackson:
The drag and drop from the DataStructure TreeView has been disabled through this patch. The  issue is DataStructureTreeView.cpp where we now put in a ProxyModel in place of the AbstractItemModel. Then when we try to extract the path we can't dynamic cast to a StandardItemModel anymore and the method early exits.

mmarineBlueQuartz:
Drag and drop has been fixed, and the GUI was updated for expand and collapse Data Structure buttons.

imikejackson:
do merge:

PR #341, opened 2019-04-18
WIP: Update IGeometryGrid API for set/get Dimension/Origin/Spacing

imikejackson:
The API is changed to use a FloatVec3Type or SizeVec3Type (both alias for SIMPLArray<T, N>) so that
the API is consistent among the various classes and easy to use within SIMPL/DREAM.3D and its plugins.
The use of the std::tuple<T> based return types is deprecated and removed.

### DIFF
--- a/Source/SIMPLib/Common/SIMPLArray.hpp
+++ b/Source/SIMPLib/Common/SIMPLArray.hpp
@@ -187,15 +187,15 @@ public:
   /**
    * @brief Converts to another container type. The output type that is being used needs to have the "push_back()" method implemented.
    *
-   *   For STL containers this includes Vector, List, Deque. QVector and QList will also work.
+   *   For STL containers this includes Vector, Deque. QVector will also work.
    */
   template <typename OutContainerType>
   OutContainerType toContainer()
   {
-    OutContainerType dest;
-    for(const auto& value : m_Array)
+    OutContainerType dest(Dimension);
+    for(typename OutContainerType::size_type i = 0; i < Dimension; i++)
     {
-      dest.push_back(value);
+      dest[i] = m_Array[i];
     }
     return dest;
   }
--- a/Source/SIMPLib/Common/SIMPLArray.hpp
+++ b/Source/SIMPLib/Common/SIMPLArray.hpp
@@ -251,6 +251,24 @@ public:
   {
     return std::make_tuple(getX(), getY());
   }
+
+  template <typename OutType>
+  IVec2<OutType> convertType()
+  {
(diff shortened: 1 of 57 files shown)


imikejackson:
All of the other repos are also updated with the same branch. I am just putting in this PR rather than a PR for all 20 repositories.

I also added a few more methods to the SIMPLArray subclasses to convert to another container type (such as QVector or std::vector) and convert to SIMPLArray of another data type.

mmarineBlueQuartz (approved):
This looks good, but it might be worth it to mention that toContainer() requires the ContainerType to allow random access in the `@brief` description.  Many STL containers do not fit this requirement.

joeykleingers (approved):
Looks good.

imikejackson:
@mmarineBlueQuartz True. Maybe I should just use the most general case for the concrete implementations. I could probably use an initializer list since we know the size of the array in those functions. I think most (all?) the STL containers and QVector can handle an initializer list.

imikejackson:
Also thought about implementing some stream operators so we can have debug output of the class.

imikejackson:
I added the requested changes to SIMPLArray, added some more constructors. Update the Doxygen for SIMPLArray to state which container types work. Updated the unit test for SIMPLArray to add the Vec4 and Vec6 cases, and add instantiations of the toContainer for std::vector, std::list, std::deque, QVector and QList.

Source/SIMPLib/Common/SIMPLArray.hpp
```diff
@@ -172,6 +184,22 @@ class SIMPLArray
...
 
+  /**
+   * @brief Converts to another container type. The output type that is being used needs to have the "push_back()" method implemented.
+   *
+   *   For STL containers this includes Vector, List, Deque. QVector and QList will also work.
+   */
+  template <typename OutContainerType>
+  OutContainerType toContainer()
+  {
+    OutContainerType dest;
```
mmarineBlueQuartz: If you don't mind losing QList (which is suggested to avoid like the plague), pass in the dimentions to `OutContainerType dest(m_Array.size());` and keep your [] operators for assignment and copying.  Every other container mentioned in the description allows that pattern, and the most likely conversions (std::vector and QVector) do so optimally.  std::list and std::deque are slower when it comes to random access but are similarly unlikely conversions from an array.

imikejackson: All requests implemented.

imikejackson:
This PR is superseded by PR #345

PR #342, opened 2019-04-18
Topic/2 d image import fix

joeykleingers:
Updating ITKImageReader and itkInPlaceImageToDream3DDataFilter classes so that they import 2D datasets with the 3rd dimension in the data container's geometry set to 0.  This is needed to be able to tell the difference between a true, flat 2D dataset and a really thin image geometry.

### DIFF
--- a/Source/SIMPLib/ITK/itkImageReaderHelper.cpp
+++ b/Source/SIMPLib/ITK/itkImageReaderHelper.cpp
@@ -253,12 +253,14 @@ ITK_IMAGE_READER_CLASS_NAME
   // Initialize torigin/tspacing/tDims since arrays are always of size 3 and ITK image may have a different size.
   FloatVec3Type torigin = {0.0f, 0.0f, 0.0f};
   FloatVec3Type tspacing = {1.0f, 1.0f, 1.0f};
-  SizeVec3Type tDims = {1, 1, 1};
+  SizeVec3Type tDims = {0, 0, 0};
+  QVector<size_t> qTdims;
   for(size_t i = 0; i < dimensions; i++)
   {
     torigin[i] = origin[i];
     tspacing[i] = spacing[i];
     tDims[i] = size[i];
+    qTdims.push_back(tDims[i]);
   }
   ImageGeom::Pointer image = ImageGeom::CreateGeometry(SIMPL::Geometry::ImageGeometry);
   image->setDimensions(tDims);
@@ -267,7 +269,6 @@ ITK_IMAGE_READER_CLASS_NAME
   container->setGeometry(image);
 
   QVector<size_t> cDims = ITKDream3DHelper::GetComponentsDimensions<TPixel>();
-  QVector<size_t> qTdims = {tDims[0], tDims[1], tDims[2]};
   AttributeMatrix::Pointer cellAttrMat = container->createNonPrereqAttributeMatrix(this, dataArrayPath.getAttributeMatrixName(), qTdims, AttributeMatrix::Type::Cell);
   if(getErrorCode() < 0)
   {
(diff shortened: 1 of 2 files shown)


mmarineBlueQuartz:
It's a bit confusing with the clang-format included, but it looks like we're continuing with ImageGeom instead of splitting 2D and 3D geometries?

joeykleingers:
Yes, we are continuing with ImageGeom until we have the time to implement the solution we talked about in Mike's office.  All 2D images will have a Z dimension of 0 when imported with ITK, which will allow our code in SIMPLVtkLib to be able to distinguish between 2D and 3D datasets.

imikejackson:
I concur that we do not have the allotment of time right now to really go down the ImageGeom2D and ImageGeom3D path even though I agree that it is probably the better solution. This will patch the issue for now and allow the feature that we are needing to be implemented for the customer.

@joeykleingers How did all the clang formatting get there? Did those lines get touched? I think a better plan to get the formatting completed would be to do a last push that is nothing more than a clang-format *after* we have accepted the PR.

Thoughts?

joeykleingers:
Both of those files were horrendously unformatted, so I purposefully made all my changes in one commit and then clang formatted in a second commit.  So you can look at the first commit to see the changes I made, and the second commit is just cleaning up the formatting.

Source/SIMPLib/ITK/itkImageReaderHelper.cpp
```diff
@@ -253,12 +244,14 @@ ::readImageOutputInformation(const DataArrayPath& dataArrayPath, typename itk::I
...
   FloatVec3Type tspacing = {1.0f, 1.0f, 1.0f};
-  SizeVec3Type tDims = {1, 1, 1};
+  SizeVec3Type tDims = {0, 0, 0};
+  QVector<size_t> qTdims;
   for(size_t i = 0; i < dimensions; i++)
   {
     torigin[i] = origin[i];
     tspacing[i] = spacing[i];
     tDims[i] = size[i];
+    qTdims.push_back(tDims[i]);
```
mmarineBlueQuartz: `push_back` isn't the fastest operation when you're increasing the size of the underlying array (should average out to O(N log(N))), so if you're importing a lot of data, you might want to use `QVector<size_t> qTdims(dimensions);` with `qTdims[i] = tDims[i];`

mmarineBlueQuartz: O(N log(N)) is a best case scenario, but it depends on implementation.  Qt collections are known to be slower than their STL counterparts, so it would be best to make it as constant-time as possible.

imikejackson: In the context of the code the largest the array will be is 3, maybe 4.

mmarineBlueQuartz: Does itkImageReaderHelper get used for stacks of images?  My guess is that the initial underlying array is size 0, so you're probably having two resizes for each image.

joeykleingers: It is only used in ITKImageReader

joeykleingers: Any other comments on this PR?  I want to get it merged in if we are ok with it.  Right now, importing 2D images (including montage tiles) is broken, and this fixes it.

joeykleingers:
Any other comments on this PR? I want to get it merged in if we are ok with it. Right now, importing 2D images (including montage tiles) is broken, and this fixes it.

imikejackson (commented):
Looks like the temp fix that we need now so do the merge. We can figure out the ImageGeom2D later this summer (Hopefully)

PR #345, opened 2019-04-24
IGeometryGrid API Update, Python Fixes, MXA Removal

imikejackson:
The API is changed to use a FloatVec3Type or SizeVec3Type (both alias for SIMPLArray<T, N>) so that
the API is consistent among the various classes and easy to use within SIMPL/DREAM.3D and its plugins.
The use of the std::tuple based return types is deprecated and removed.

The Python script files were completely reorganized to put the scripts with their plugins. Most of
the scripts were tested on a macOS machine and added to the ctest suite of unit tests. For each script
a bash script is created that drives the test. The Python tests are dependant on an Anaconda3 installation
being used. A few of the Python tests are still failing but those will be fixed eventually

The source code was searched for the old and deprecated MXA strings and classes and the old code was
either removed or renamed to SIMPL.

Signed-off-by: Michael Jackson mike.jackson@bluequartz.net


Source/SIMPLib/Common/SIMPLArray.hpp
```diff
@@ -244,21 +297,39 @@ class IVec2 : public SIMPLArray<T, 2>
...
+    (*this)[0] = x;
+    (*this)[1] = y;
+  }
+
+  /**
+   * @brief values Converts the internal data structure to a std::tuple<T, T, T>
    * @return
    */
-  std::tuple<T, T> toTuple() const
+  std::tuple<T, T> values() const
```
mmarineBlueQuartz: toTuple was probably the better name.  Asking an array for its values and getting a tuple is not my expected result if I just saw a list of auto-complete names in an IDE.  I would expect a pointer or vector to come back from just this method name.

Source/SIMPLib/Common/SIMPLArray.hpp
```diff
@@ -428,13 +568,165 @@ template <typename T> class IVec4 : public SIMPLArray<T, 4>
...
+   */
+  inline void setValues(const T& x, const T& y, const T& z, const T& w)
+  {
+    (*this)[0] = x;
+    (*this)[1] = y;
+    (*this)[2] = z;
+    (*this)[3] = w;
+  }
+
+  std::tuple<T, T, T, T> values() const
```
mmarineBlueQuartz: Missing Doxygen documentation

mmarineBlueQuartz: Check your other IVec#::values() as well.

Source/SIMPLib/Common/SIMPLArray.hpp
imikejackson: Somewhat agree. I did this change because the Python bindings need a method to return a tuple but inside of Python it is a list which I thought would be confusing inside of python. I wasn't super sure about the name but nothing else really worked. Let me take a look at the Python binding code and see if something can be done to have it both ways.

Source/SIMPLib/Geometry/ImageGeom.cpp
```diff
@@ -235,7 +234,7 @@ class FindImageDerivativesImpl
   }
 #endif
 
-  void computeIndices(int32_t differenceType, int32_t directionType, size_t& index1, size_t& index2, size_t dims[3], size_t x, size_t y, size_t z, double xp[3], double xm[3]) const
+  void computeIndices(int32_t differenceType, int32_t directionType, size_t& index1, size_t& index2, const size_t dims[3], size_t x, size_t y, size_t z, double xp[3], double xm[3]) const
```
mmarineBlueQuartz, line 237: Because dims[3] has a const, I'm going to ask for confirmation that xm[3] and xp[3] are expected to have their values updated and "returned" by this method.  Otherwise, the inconsistency is misleading.

sdonegan: I can confirm the xm and xp are modified in this function; they are ultimately passed into a call to getCoords() which will set their values to the proper (x,y,z) coordinates depending on the derivative stencil.

Source/SIMPLib/Common/SIMPLArray.hpp
imikejackson: @mmarineBlueQuartz I fixed this in the latest push

imikejackson:
All the requested changes have been implemented. All repositories are updated. All unit tests are passing on all platforms.

imikejackson:
Even builds on Linux when a compatible Anaconda is installed....

PR #347, opened 2019-05-02
All the IGeometry subclasses are updated to use size_t instead of int64_t

imikejackson:
When asking for the number of items/elements in a geometry the type should be a size_t
since those values are natively stored as size_t in the underlying classes.


Source/SIMPLib/Geometry/SharedEdgeOps.cpp
```diff
@@ -78,7 +78,7 @@ SharedEdgeList::Pointer GEOM_CLASS_NAME::getEdges()
 // -----------------------------------------------------------------------------
 //
 // -----------------------------------------------------------------------------
-void GEOM_CLASS_NAME::setVertsAtEdge(int64_t edgeId, int64_t verts[2])
+void GEOM_CLASS_NAME::setVertsAtEdge(size_t edgeId, int64_t verts[2])
```
mmarineBlueQuartz: I believe verts[2] should be size_t as well, although that will require additional changes.

mmarineBlueQuartz (changes requested):
Make sure actual vertex IDs are updated in geometry and Shape**Ops classes as well.  Having size_t counts is of reduced usefulness when the actual IDs to lookup or assign are themselves signed values.  I particularly noticed this with edges, but every time there were multiple ID parameters, only the first one got updated.

You might want to consider adding VertexID typedef to further document what the value is and differentiate the multiple uses of the `vert` parameter name.

Source/SIMPLib/Geometry/SharedEdgeOps.cpp
imikejackson: The actual value of the vertex is an int64_t, the index into the array is a size_t.

mmarineBlueQuartz: Correct me if I'm wrong, but the value of the vertex is its ID, right?  I'm confused why only some IDs are being updated and some are not.

imikejackson: Let me go back and work my way through it again. I'm just not really sure why int64_t was used for ids.

mmarineBlueQuartz (approved):
This looks good, and I'd be happy with it getting merged in as is.

There is an improvement that could be made, though.  A quarter of the way through, use of the MeshIndexType type alias fell off in favor of size_t.  I do not see that alias being changed, but it would be better to keep using the ID alias unless referring to an actual size or count.  This would give more context to the parameters without having to rename them or add comments.

imikejackson:
I think it was the other way around. At some point I started updating variables that were specifically being used to index through a mesh geometry. I can try and find more of those instances. Not every occurrence of size_t is that type of looping so I hesitate to make a mass change.

PR #348, opened 2019-05-07
The DataStructureWidget no longer expands all items on refresh

mmarineBlueQuartz:
* Refreshing the data in DataStructureWidget no longer expands all items.
* Adding child nodes to the DataStructureWidget now expands the parent if the parent did not have any children previously.
* Removed commented out styling for the DataStructure.  This has been handled by the stylesheets for some time and is unlikely to ever return to DataStructureWidget.


imikejackson (commented):
Looks good. Merge it in.

PR #350, opened 2019-05-14
Added header file that allows simpler coding for parallel algorithms

mmarineBlueQuartz:
* Added ParallelDataAlgorithm and ParallelTaskAlgorith for running differing types of algorithms.  Alternate versions are available for uses where SIMPL_USE_PARALLEL_ALGORITHMS is not defined.  This avoids the requirement of repetitive checks for SIMPL_USE_PARALLEL_ALGORITHMS and varying implementations.
* This header also includes the required tbb includes if SIMPL_USE_PARALLEL_ALGORITHMS is defined.
* Updated SIMPLib parallel algorithms to use these new classes.


imikejackson (changes requested):
Overall looks really great. The copyright block needs to be adjusted to reflect *only* the current DoD contract number of FA8650-15-D-5231

The updated codes are much cleaner now.

imikejackson:
Do all the unit tests pass? Are there changes that would interfere with the Python bindings?

imikejackson (commented):
Over all good design. See other questions in the comments.

imikejackson (commented):
Since SIMPLRange , SIMPLRange2D and SIMPLRange3D are *not* templated classes I think we should split the definition and declaration into .cpp and .h files?

imikejackson:
Also some misc style fixes:

+ Keep the SIMPL headers together up in the include section of the sources so clang format can do it's sorting job.
+ The TBB headers should be included *before* the SIMPL headers since SIMPL depends on TBB (assuming the developer elected to compile against TBB)

mmarineBlueQuartz:
SIMPL includes cannot be placed after the TBB includes because the TBB include guards are dependent on SIMPLib.h to define whether or not TBB is available.  This is consistent with develop, only develop hides this behind the separation of header and source files where TBB is only required in the source and SIMPLib.h is included in the header.  That is not an option when a constructor directly accepts TBB classes if and only if SIMPL is configured to use TBB.

imikejackson:
Good Explanation. Can you place that explanation above the includes in question. That ensures that no one else decides to reorganize those includes. Also "guard" the includes with a // clang-format off and // clang-format on so that clang format does not touch the include order.

imikejackson (approved):

PR #351, opened 2019-05-17
Fixing bug where turning off SIMPL_USE_ITK would still include the ITK directory in SIMPLib.

joeykleingers:


mmarineBlueQuartz (approved):
This looks good to me

PR #352, opened 2019-05-29
Feature/montage selection

mmarineBlueQuartz:
* Added MontageSelection class to contain values required for selecting a range of montage tiles.
* Added MontageSelectionFilterParameter to wrap MontageSelection and allow the creation of the corresponding MontageSelectionWidget.  This allows the user to select a range of DataContainers for a montage based on start and end values for row, a padding amount for digits, as well as prefix and suffix values surrounding the DataContainer name.
* The resulting DataContainers matching the MontageSelection are displayed in MontageSelectionWidget in a QTableWidget.  DataContainers that exist in the filter's DataContainerArray are marked with a green bullet icon.  DataContainers that could not be found are marked with a red bullet icon.


imikejackson (changes requested):
You don't need (and we don't want) the FP reading to be inside of AbstractFilterParametersReader since that class reads/writes to HDF5 files which are not supported anymore. We store the complete json inside the HDF5 file since a few versions of DREAM.3D

Need the Prime Contract number for NRL that this work was done under for new source files (missing in one of the .cpp files)

In the filter parameter itself the constructor should be something like:
        const QString& humanLabel, const QString& propertyName, const IntVec3Type& defaultValue, Category category, const SetterCallbackType& setterCallback, const GetterCallbackType& getterCallback, int groupIndex)

Will need to expose MontageSelection to the Pybind11 system so that a user on the python side of things can use any filter that requires a MontageSelection object as a filter parameter

imikejackson:
Take a look at DynamicTableData to see how we exposed it for Python.

mmarineBlueQuartz:
The requested changes have been made.

PyBind11 tests fail even on develop, so I'm unsure if this is another issue between building Python bindings on Windows or if it's just my system.

PR #353, opened 2019-05-29
Feature/IntVec2FilterParameter

mmarineBlueQuartz:
* Created IntVec2FilterParameter and corresponding IntVec2Widget based on IntVec3FilterParameter and IntVec3Widget.
* Added IntVec2Type in the FilterParameter readers and writers.
* Added IntVec2FilterParameter to GenericExample.

### DIFF
--- a/Source/SIMPLib/FilterParameters/AbstractFilterParametersReader.cpp
+++ b/Source/SIMPLib/FilterParameters/AbstractFilterParametersReader.cpp
@@ -348,18 +348,6 @@ QVector<double> AbstractFilterParametersReader::readArray(const QString name, QV
 // -----------------------------------------------------------------------------
 //
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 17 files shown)


mmarineBlueQuartz:
No other filters currently use IntVec2FilterParameter at the moment.

imikejackson (commented):
Same as other PR: We do not need the ability to R/W the FP from the HDF5 file. Nobody would have a requirement for this since this is a new FP. Please remove these additions to the various classes.

Adjust the Prime Contract number that these sources were developed under in the license section.

IntVec2Widget.h The order of includes should be C/C++, external libraries, SIMPLib, SVwidgetsLib

mmarineBlueQuartz:
Changes have been made, and the unit tests pass.

imikejackson (approved):

PR #359, opened 2019-06-07
Various small code changes and cleanups to allow deeper refactoring

imikejackson:
These commits modify source codes that would allow the mass removal of the use of the SIMPL_* macros.


Source/SIMPLib/Messages/AbstractStatusMessage.h
```diff
@@ -50,33 +49,25 @@ class SIMPLib_EXPORT AbstractStatusMessage : public AbstractMessage
...
-  virtual ~AbstractStatusMessage()
-  {
-  }
+  ~AbstractStatusMessage() override;
 
   /**
    * @brief This method creates and returns a string for status messages
    */
-  virtual QString generateMessageString() const = 0;
+  QString generateMessageString() const override = 0;
```
mmarineBlueQuartz: It would probably be best to just remove cases where we override a pure virtual method with a pure virtual method.  There is nothing being added, and it seems to imply that the base actually has an implementation that needs to be ignored or overridden in derived classes.

Source/SIMPLib/Geometry/ITransformContainer.cpp
```diff
@@ -0,0 +1,37 @@
...
+*
+* The code contained herein was partially funded by the followig contracts:
+*    United States Air Force Prime Contract FA8650-15-D-5231
+*
+* ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ */
+#include "ITransformContainer.h"
+
+ITransformContainer::ITransformContainer() = default;
+
+ITransformContainer::~ITransformContainer() = default;
```
mmarineBlueQuartz: Missing a new line at end of file.

mmarineBlueQuartz (changes requested):
Remove the `override = 0;` method declarations unless they are overriding actual implementations.  It muddies the intent without adding anything of value.  I could assume the virtual methods had a purpose without checking the base classes, but that becomes harder when using override.

Otherwise, this looks good.

mmarineBlueQuartz (changes requested):
Pure virtual overrides exist in each of the abstract message types (AbstractErrorMessage, AbstractProgressMessage, and AbstractWarningMessage).  Only AbstractStatusMessage was updated.

imikejackson:
I rebased the branch on the latest develop pushes. The requested changes were made to the 4 files in question. I addition the TYPE macro was changed to the proper type to ensure the override keyword was used correctly in the getNameOfClass() method which was overridden in the macro.

mmarineBlueQuartz (approved):
Looks good to me

PR #360, opened 2019-06-10
WIP changes to how numpy arrays are converted to/from DataArrays

sdonegan:
This PR changes how SIMPL interacts with numpy arrays.  The following changes are included:

- Changed the standard constructor when creating a SIMPL DataArray object that wraps an underlying numpy array.  Now, the constructor only takes as arguments the numpy array, a string name, and a flag for whether the new DataArray will own the data.  The major change here is how the tuples are component dimensions are inferred from the numpy array shape.  The first axis of the numpy array (slowest-moving in C-order) is now considered the number of tuples, while the following axes (faster-moving in C-ordering) determine the component dimensions.  

Consider a numpy array with shape `[10, 3]`, something that could represent 10 rgb values or 10 Euler angles.  The number of tuples in this case is 10 and the component dimensions will be `[3]`.  For another example, consider a numpy array with shape `[1000, 80, 60]`.  Here, the number of tuples would be 1000 while the component dimensions would be `[80, 60]`.  Additionally, scalar (0-dim) numpy arrays are explicitly handled.

- Added a `.def_buffer` lambda that provides buffer access to the underlying DataArray.  This allows direct access of the SIMPL DataArray object as a numpy array.  The convention for inferring numpy array shape is the same as above; for example, a DataArray with 10 tuples and component dimensions `[3]` becomes a numpy array with shape `(10, 3)`.  Note that there is an ambiguity when dealing with scalar DataArrays (i.e., component dimensions of `[1]`).  In this case, both `(num_tuples,)` and `(num_tuples, 1)` are valid numpy array shapes.  The chosen convention is to squeeze the 1-dim axis in the numpy array; i.e., the returned numpy array shape will be `(num_tuples,)`.  To convert a DataArray to a numpy array, the `np.asarray()` function should be used.

The following functions in simpl_helpers.py will need to be updated for the new syntax: `CreateDataArray` and `ConvertToDataArray`.  Also, a helper function that converts a DataArray to a numpy array needs to be added.

Now, some example code showing round-tripping numpy and DataArray objects:

```python
num_array = np.arange(30, dtype=np.float32).reshape([5, 3, 2])
z = np.asarray(num_array)
if not z.flags.contiguous:
    z = np.ascontiguousarray(z)

print("--------------- z ---------------")
print(z.shape)
print(z)

array = simpl.FloatArrayType(z, "test", False)

roundtrip = np.asarray(array)

print("--------------- roundtrip ---------------")
print(roundtrip.shape)
print(roundtrip)
```

Output of the above code:

```
--------------- z ---------------
(5, 3, 2)
[[[ 0.  1.]
  [ 2.  3.]
  [ 4.  5.]]

 [[ 6.  7.]
  [ 8.  9.]
  [10. 11.]]

 [[12. 13.]
  [14. 15.]
  [16. 17.]]

 [[18. 19.]
  [20. 21.]
  [22. 23.]]

 [[24. 25.]
  [26. 27.]
  [28. 29.]]]
--------------- roundtrip ---------------
(5, 3, 2)
[[[ 0.  1.]
  [ 2.  3.]
  [ 4.  5.]]

 [[ 6.  7.]
  [ 8.  9.]
  [10. 11.]]

 [[12. 13.]
  [14. 15.]
  [16. 17.]]

 [[18. 19.]
  [20. 21.]
  [22. 23.]]

 [[24. 25.]
  [26. 27.]
  [28. 29.]]]
```

### DIFF
--- a/Wrapping/Python/Pybind11/Templates/SIMPLModuleCodeTemplate.in.cpp
+++ b/Wrapping/Python/Pybind11/Templates/SIMPLModuleCodeTemplate.in.cpp
@@ -82,24 +82,89 @@ template <typename T> using PySharedPtrClass = py::class_<T, std::shared_ptr<T>>
         .def(py::init([](T* ptr, size_t numElements, std::vector<size_t> cDims, QString name, bool ownsData) {                                                                                         \
           return DataArrayType::WrapPointer(ptr, numElements, QVector<size_t>::fromStdVector(cDims), name, ownsData);                                                                                  \
         }))                                                                                                                                                                                            \
-        .def(py::init([](py::array_t<T, py::array::c_style> b, std::vector<size_t> cDims, QString name, bool ownsData) {                                                                               \
-          ssize_t numElements = 1;                                                                                                                                                                     \
-          ssize_t nDims = b.ndim();                                                                                                                                                                    \
-          for(ssize_t e = 0; e < nDims; e++)                                                                                                                                                           \
+        .def(py::init([](py::array_t<T, py::array::c_style> b, QString name, bool ownsData) {                                                                                                          \
+          py::buffer_info buf = b.request();                                                                                                                                                           \
+          std::vector<ssize_t> shape = buf.shape;                                                                                                                                                      \
+          ssize_t ndim = buf.ndim;                                                                                                                                                                     \
+          size_t numTuples = 0;                                                                                                                                                                        \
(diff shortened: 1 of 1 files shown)


tplatt77 (approved):
Looks good to me

imikejackson:
Codes based off of this source have been merged into develop. Closing PR

PR #362, opened 2019-06-11
WIP: Changing the IDataArray and DataArray API to use STL containers.

imikejackson:
* Predominant change is to use std::vector<size_t> for the component dimensions argument.
* Other minor clang-tidy warning cleanup if deemed appropriate.

### DIFF
--- a/Source/SIMPLib/Common/TemplateHelpers.cpp
+++ b/Source/SIMPLib/Common/TemplateHelpers.cpp
@@ -37,7 +37,8 @@
 using namespace TemplateHelpers;
 
 // -----------------------------------------------------------------------------
-IDataArrayWkPtr CreateNonPrereqArrayFromArrayType::operator()(AbstractFilter* f, const DataArrayPath& arrayPath, const QVector<size_t>& compDims, const IDataArrayShPtr& sourceArrayType, RenameDataPath::DataID_t id)
+IDataArrayWkPtr CreateNonPrereqArrayFromArrayType::operator()(AbstractFilter* f, const DataArrayPath& arrayPath, const std::vector<size_t>& compDims, const IDataArrayShPtr& sourceArrayType,
+                                                              RenameDataPath::DataID_t id)
 {
 
   IDataArrayShPtr ptr = IDataArray::NullPointer();
@@ -99,7 +100,8 @@ IDataArrayWkPtr CreateNonPrereqArrayFromArrayType::operator()(AbstractFilter* f,
 }
 
 // -----------------------------------------------------------------------------
-IDataArrayWkPtr CreateNonPrereqArrayFromTypeEnum::operator()(AbstractFilter* f, const DataArrayPath& arrayPath, const QVector<size_t>& compDims, int arrayType, double initValue, RenameDataPath::DataID_t id)
+IDataArrayWkPtr CreateNonPrereqArrayFromTypeEnum::operator()(AbstractFilter* f, const DataArrayPath& arrayPath, const std::vector<size_t>& compDims, int arrayType, double initValue,
+                                                             RenameDataPath::DataID_t id)
 {
   IDataArrayShPtr ptr = IDataArray::NullPointer();
 
@@ -150,16 +152,16 @@ IDataArrayWkPtr CreateNonPrereqArrayFromTypeEnum::operator()(AbstractFilter* f, 
 }
 
 // -----------------------------------------------------------------------------
-IDataArrayShPtr CreateArrayFromArrayType::operator()(AbstractFilter* f, const size_t& numTuples, const QVector<size_t>& compDims, const QString& arrayName, bool allocate,
+IDataArrayShPtr CreateArrayFromArrayType::operator()(AbstractFilter* f, const size_t& numTuples, const std::vector<size_t>& compDims, const QString& arrayName, bool allocate,
                                                      const IDataArrayShPtr& sourceArrayType)
 {
   CreateArrayFromArrayType classInstance;
-  QVector<size_t> tupleDims(1, numTuples);
+  std::vector<size_t> tupleDims = {numTuples};
   return classInstance(f, tupleDims, compDims, arrayName, allocate, sourceArrayType);
 }
 
 // -----------------------------------------------------------------------------
-IDataArrayShPtr CreateArrayFromArrayType::operator()(AbstractFilter* f, const QVector<size_t>& tupleDims, const QVector<size_t>& compDims, const QString& arrayName, bool allocate,
+IDataArrayShPtr CreateArrayFromArrayType::operator()(AbstractFilter* f, const std::vector<size_t>& tupleDims, const std::vector<size_t>& compDims, const QString& arrayName, bool allocate,
                                                      const IDataArrayShPtr& sourceArrayType)
 {
   IDataArrayShPtr ptr = IDataArray::NullPointer();
@@ -221,15 +223,16 @@ IDataArrayShPtr CreateArrayFromArrayType::operator()(AbstractFilter* f, const QV
 }
 
 // -----------------------------------------------------------------------------
-IDataArrayShPtr CreateArrayFromType::operator()(AbstractFilter* f, const size_t& numTuples, const QVector<size_t>& compDims, const QString& arrayName, bool allocate, const QString& type)
+IDataArrayShPtr CreateArrayFromType::operator()(AbstractFilter* f, const size_t& numTuples, const std::vector<size_t>& compDims, const QString& arrayName, bool allocate, const QString& type)
 {
   CreateArrayFromType classInstance;
-  QVector<size_t> tupleDims(1, numTuples);
+  std::vector<size_t> tupleDims(1, numTuples);
   return classInstance(f, tupleDims, compDims, arrayName, allocate, type);
 }
 
 // -----------------------------------------------------------------------------
-IDataArrayShPtr CreateArrayFromType::operator()(AbstractFilter* f, const QVector<size_t>& tupleDims, const QVector<size_t>& compDims, const QString& arrayName, bool allocate, const QString& type)
+IDataArrayShPtr CreateArrayFromType::operator()(AbstractFilter* f, const std::vector<size_t>& tupleDims, const std::vector<size_t>& compDims, const QString& arrayName, bool allocate,
+                                                const QString& type)
 {
   IDataArrayShPtr ptr = IDataArray::NullPointer();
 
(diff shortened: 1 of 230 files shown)


imikejackson:
Please have a look at the DataArray API for constructors. They have all been moved closer to the top of the source file and grouped together. There was a protected constructor that is now public. I was striving to make the API consistent among the constructors so that each constructor add another argument to the last in an unambiguous way.

Source/SVWidgetsLib/FilterParameterWidgets/FilterParameterWidgetUtils.hpp
```diff
@@ -216,7 +216,19 @@ class FilterParameterWidgetUtils
       }
     }
 
-
+    // -----------------------------------------------------------------------------
+    template <typename T>
+    static bool VectorContains(const std::vector<std::vector<T>>& container, const std::vector<T>& comparison)
+    {
+      for(const auto& value : container)
```
mmarineBlueQuartz: return std::find(...) != container.end()

imikejackson: Didn't know about that one... I thought it was strange there was no way to "find" or "contains" something.

Source/SIMPLib/FilterParameters/MultiAttributeMatrixSelectionFilterParameter.h
```diff
@@ -79,7 +79,7 @@ class SIMPLib_EXPORT MultiAttributeMatrixSelectionFilterParameter : public Filte
       IGeometry::Types dcGeometryTypes;
       AttributeMatrix::Types amTypes;
       QVector<QString> daTypes;
-      QVector< QVector<size_t> > componentDimensions;
+      QVector< std::vector<size_t> > componentDimensions;
```
mmarineBlueQuartz, line 82: Is this supposed to still be a QVector, or did the space throw off text replacement?  Same issue occurs in MultiDataArraySelectionFilterParameter.

Source/SVWidgetsLib/FilterParameterWidgets/FilterParameterWidgetUtils.hpp
mmarineBlueQuartz: Anything with STL iterator support can be used in std::find, std::find_if, and many other interesting functions in `<algorithm>`.  The reason the type doesn't have it is because the method is already generically available.

Source/SIMPLib/FilterParameters/MultiAttributeMatrixSelectionFilterParameter.h
imikejackson: Looks like we are still using QVector. I can correct them which will have a rolling effect through the rest of the code base. If we want to keep going down this path I will get those updated also. Which also starts bringing up the question of do we want to just start removing all instance of QVector? like for QVector<QString> and things like that?

mmarineBlueQuartz: Most cases of QVector were replaced, so I mostly wanted to make sure that the few that were left were intentional or at least known going forward.  I can see cases where handling Qt classes would make sense to use QVector due to it being a return type or parameter in Qt.

Using QVector over standard types feels like we're using it just to use it.  I'm not sure about the worth of forcing the change just to force the change, but time aside, I lean towards using STL collections over Qt collections.

Source/SIMPLib/Common/TemplateHelpers.cpp
```diff
@@ -150,16 +152,16 @@ IDataArrayWkPtr CreateNonPrereqArrayFromTypeEnum::operator()(AbstractFilter* f,
 }
 
 // -----------------------------------------------------------------------------
-IDataArrayShPtr CreateArrayFromArrayType::operator()(AbstractFilter* f, const size_t& numTuples, const QVector<size_t>& compDims, const QString& arrayName, bool allocate,
+IDataArrayShPtr CreateArrayFromArrayType::operator()(AbstractFilter* f, const size_t& numTuples, const std::vector<size_t>& compDims, const QString& arrayName, bool allocate,
                                                      const IDataArrayShPtr& sourceArrayType)
 {
   CreateArrayFromArrayType classInstance;
-  QVector<size_t> tupleDims(1, numTuples);
+  std::vector<size_t> tupleDims = {numTuples};
```
mmarineBlueQuartz, line 159: I don't believe the `=` is necessary.  Upon checking the constructors on cppreference, they choose not to include `=`.

Source/SIMPLib/CoreFilters/GenerateColorTable.cpp
```diff
@@ -63,13 +63,13 @@ template <typename T>
...
 {
 public:
-  GenerateColorTableImpl(typename DataArray<T>::Pointer arrayPtr, QVector<float> binPoints,
-                         std::vector<std::vector<double> > controlPoints, int numControlColors, UInt8ArrayType::Pointer colorArray) :
-    m_ArrayPtr(arrayPtr),
-    m_BinPoints(binPoints),
-    m_NumControlColors(numControlColors),
-    m_ControlPoints(controlPoints),
-    m_ColorArray(colorArray)
+  GenerateColorTableImpl(typename DataArray<T>::Pointer arrayPtr, std::vector<float>& binPoints, std::vector<std::vector<double>>& controlPoints, int numControlColors,
```
mmarineBlueQuartz: `binPoints` changed from pass by value to pass by reference, and it's not const.

mmarineBlueQuartz: Watch `controlPoints` as well.

Source/SIMPLib/CoreFilters/ImportAsciDataArray.cpp
```diff
@@ -428,14 +428,14 @@ void ImportAsciDataArray::dataCheck()
     return;
   }
 
-  QVector<size_t> tDims = attrMat->getTupleDimensions();
+  std::vector<size_t> tDims = attrMat->getTupleDimensions();
   size_t totalDim = 1;
   for(int i = 0; i < tDims.size(); i++)
```
mmarineBlueQuartz: Side note: this should give a signed / unsigned conversion warning now.  `size()` really should (and now will) return an unsigned value.

Source/SIMPLib/FilterParameters/DataArraySelectionFilterParameter.cpp
```diff
@@ -93,7 +88,7 @@ DataArraySelectionFilterParameter::RequirementType DataArraySelectionFilterParam
                                                                                                                 size_t allowedCompDim,
                                                                                                                 AttributeMatrix::Category attributeMatrixCategory)
 {
-  typedef QVector<size_t> QVectorOfSizeType;
+  using QVectorOfSizeType = std::vector<size_t>;
```
mmarineBlueQuartz: This is a misleading and bloated type name.  If the name can be changed without breaking too much, consider using SizeVecType instead.
```suggestion
  using SizeVecType = std::vector<size_t>;
```

Source/SIMPLib/FilterParameters/DataArraySelectionFilterParameter.cpp
```diff
@@ -141,15 +136,15 @@ DataArraySelectionFilterParameter::RequirementType DataArraySelectionFilterParam
                                                                                                         AttributeMatrix::Type attributeMatrixType,
                                                                                                         IGeometry::Type geometryType)
 {
-  typedef QVector<size_t> QVectorOfSizeType;
+  using QVectorOfSizeType = std::vector<size_t>;
```
mmarineBlueQuartz: Again, this name is not accurate anymore.

Source/SIMPLib/FilterParameters/MultiAttributeMatrixSelectionFilterParameter.h
mmarineBlueQuartz: After seeing what feels like a dozen `QVector<QVector<T>>` converted to `std::vector<std::vector<T>>` and not `QVector<std::vector<T>>`, this feels wrong.

Source/SVWidgetsLib/Widgets/DataArrayPathSelectionWidget.cpp
```diff
@@ -271,6 +271,19 @@ bool DataArrayPathSelectionWidget::CheckPathRequirements(AbstractFilter* filter,
   return true;
 }
 
+template <typename T>
+bool vectorContains(const std::vector<std::vector<T>>& container, const std::vector<T>& comparison)
+{
+  for(const auto& value : container)
```
mmarineBlueQuartz, line 279: - std::find
- This isn't the first of these contains helper methods. Consider moving them to a utilities .hpp
- For all intents and purposes, `std::vector<T>` IS `T`.  You can remove one of the `std::vector`s in container.

Wrapping/Python/Pybind11/Templates/SIMPLModuleCodeTemplate.in.cpp
```diff
@@ -80,7 +80,7 @@ template <typename T> using PySharedPtrClass = py::class_<T, std::shared_ptr<T>>
     PySharedPtrClass<DataArrayType> instance(m, #NAME, parent, py::buffer_protocol());                                                                                                                 \
     instance.def(py::init([](size_t numElements, QString name, bool allocate) { return DataArrayType::CreateArray(numElements, name, allocate); }))                                                    \
         .def(py::init([](T* ptr, size_t numElements, std::vector<size_t> cDims, QString name, bool ownsData) {                                                                                         \
-          return DataArrayType::WrapPointer(ptr, numElements, QVector<size_t>::fromStdVector(cDims), name, ownsData);                                                                                  \
+          return DataArrayType::WrapPointer(ptr, numElements, std::vector<size_t>::fromStdVector(cDims), name, ownsData);                                                                                  \
```
mmarineBlueQuartz:
```suggestion
          return DataArrayType::WrapPointer(ptr, numElements, cDims, name, ownsData);                                                                                  \
```

Wrapping/Python/Pybind11/Templates/SIMPLModuleCodeTemplate.in.cpp
```diff
@@ -90,7 +90,7 @@ template <typename T> using PySharedPtrClass = py::class_<T, std::shared_ptr<T>>
             numElements *= b.shape(e);                                                                                                                                                                 \
           }                                                                                                                                                                                            \
           numElements /= cDims[0];                                                                                                                                                                     \
-          return DataArrayType::WrapPointer(reinterpret_cast<T*>(b.mutable_data(0)), static_cast<size_t>(numElements), QVector<size_t>::fromStdVector(cDims), name, ownsData);                         \
+          return DataArrayType::WrapPointer(reinterpret_cast<T*>(b.mutable_data(0)), static_cast<size_t>(numElements), std::vector<size_t>::fromStdVector(cDims), name, ownsData);                         \
```
mmarineBlueQuartz:
```suggestion
          return DataArrayType::WrapPointer(reinterpret_cast<T*>(b.mutable_data(0)), static_cast<size_t>(numElements), cDims, name, ownsData);                         \
```

Source/SIMPLib/FilterParameters/MultiAttributeMatrixSelectionFilterParameter.h
mmarineBlueQuartz: Also, all `QVector<std::vector<T>>` uses I saw after this seem to come from `[DataStructure]SelectionFilterParameter` classes rather than somewhere else.

Source/SIMPLib/FilterParameters/DataArraySelectionFilterParameter.cpp
mmarineBlueQuartz: I only found this alias in .cpp files, so it is unlikely that a mass search and replace for using and typedef versions will break anything.

mmarineBlueQuartz (approved):
This looks good to me.

PR #364, opened 2019-06-14
Fix crash when reading prior .dream3d files with mesh based geometry

### LINKED ISSUE #363: Reading mesh type geometries from .dream3d files from versions prior to 6.6 results in crash
Prior versions used a Int64_t for the mesh index. Newer versions use size_t, aka, unsigned long long int. programs will crash when trying to read a mesh index dataset from the hdf5 file due to a dynamic cast that fails.

imikejackson:
The crash is due to a bad cast from Int64 array to size_t array.
We now try to first read as the Int64 array and reinterpret cast to size_t. If that fails
then we read as a size_t (unsigned long long int) and return the result of that
operation.

fixes #363. closes #363


mmarineBlueQuartz (approved):
This looks good to me.

imikejackson:
Unit tests are running on Windows 10. All passed.

PR #367, opened 2019-06-20
Remove use of SIMPL_* macros and replace with modern C++ code

imikejackson:
A number of things were done with this branch.

* All uses of SIMPL_* macros were replaced with real code
* #ifdef around the pybind11 sections
* Pybind11 macros move to their own header
* #include DataContainerArray.h was replace with a forward declaration in AbstractFilter
* Filters and other classes updated with needed include directives.


mmarineBlueQuartz:
Assuming the PyBind macros don't force the issue, it might be good to search and replace `const T&` with `T` for the following types:

- int
- float
- double
- bool
- char
- ArrayCalculator::AngleUnit
- SIMPL::ScalarTypes::Type
- SIMPL::NumericTypes::Type

Source/SIMPLib/CoreFilters/Testing/Cxx/ArrayCalculatorTest.cpp
```diff
@@ -53,13 +53,32 @@
...
 class DummyObserver : public Observer
 {
 public:
   DummyObserver() = default;
   ~DummyObserver() override = default;
-  SIMPL_TYPE_MACRO(DummyObserver)
+  /**
+   * @brief Returns the name of the class for DummyObserver
+   */
+  /**
```
mmarineBlueQuartz, line 65: Something odd happened here, and you've got duplicate comment blocks for this method.

Source/SIMPLib/CoreFilters/Testing/Cxx/MultiThresholdObjects2Test.cpp
```diff
@@ -49,13 +48,32 @@
...
 
 class MultiThresholdObjects2Test
 {
 public:
   MultiThresholdObjects2Test() = default;
   virtual ~MultiThresholdObjects2Test() = default;
-  SIMPL_TYPE_MACRO(MultiThresholdObjects2Test)
+  /**
+   * @brief Returns the name of the class for MultiThresholdObjects2Test
+   */
```
mmarineBlueQuartz, line 62: Duplicate method descriptions again.

Source/SIMPLib/CoreFilters/Testing/Cxx/MultiThresholdObjectsTest.cpp
```diff
@@ -47,13 +46,32 @@
...
 
 class MultiThresholdObjectsTest
 {
 public:
   MultiThresholdObjectsTest() = default;
   virtual ~MultiThresholdObjectsTest() = default;
-  SIMPL_TYPE_MACRO(MultiThresholdObjectsTest)
+  /**
+   * @brief Returns the name of the class for MultiThresholdObjectsTest
+   */
```
mmarineBlueQuartz, line 60: And another duplicate description.

mmarineBlueQuartz: This seems to be the case for much of the Testing/Cxx directory.  Consider this a recurring comment rather than making note of all the rest of the test classes.

Source/SIMPLib/CoreFilters/util/ABSOperator.cpp
```diff
@@ -58,7 +58,11 @@ ABSOperator::~ABSOperator() = default;
 // -----------------------------------------------------------------------------
 //
 // -----------------------------------------------------------------------------
-void ABSOperator::calculate(AbstractFilter* filter, DataArrayPath calculatedArrayPath, QStack<ICalculatorArray::Pointer>& executionStack)
+void ABSOperator::calculate(AbstractFilter* filter, DataArrayPath calculatedArrayPath,
+                            QStack<ICalculatorArray::Pointer>& executionStack){CREATE_NEW_ARRAY_STANDARD_UNARY(filter, calculatedArrayPath, executionStack, fabs)}
```
mmarineBlueQuartz: I don't think clang-format did this, but if it did, something needs to be tweaked.  I more than half-expected this line to be a lambda and was wondering where the `[=]` and closing parenthesis were.

mmarineBlueQuartz: This seems to be a recurring problem in ArrayCalculator's operators for this same method.

Source/SIMPLib/CoreFilters/util/ICalculatorArray.h
```diff
@@ -38,6 +38,10 @@
 
 #include "SIMPLib/DataArrays/DataArray.hpp"
 #include "SIMPLib/SIMPLib.h"
+#include "SIMPLib/DataArrays/DataArray.hpp"
```
mmarineBlueQuartz, line 42: Duplicate DataArray includes

Source/SIMPLib/TestFilters/FilterGroup00.cpp
```diff
@@ -187,3 +189,32 @@ const QString TESTCLASSNAME::getHumanLabel() const
...
+// -----------------------------------------------------------------------------
+FilterGroup00::Pointer FilterGroup00::NullPointer()
+{
+  return Pointer(static_cast<Self*>(nullptr));
+}
+
+// -----------------------------------------------------------------------------
+std::shared_ptr<TESTCLASSNAME> FilterGroup00::New()
+{
+  struct make_shared_enabler : public TESTCLASSNAME
```
mmarineBlueQuartz: Use of TESTCLASSNAME instead of FilterGroup00 implies that FilterGroup00 may have been mass replaced by accident.  I strongly suggest looking at the .h / .cpp pair and confirm that everything is as expected.

Source/SVWidgetsLib/FilterParameterWidgets/DataContainerArrayProxyWidget.cpp
```diff
@@ -37,10 +37,13 @@
...
 #include <cassert>
 
-
 #include <QtGui/QBrush>
 #include <QtGui/QColor>
 #include <QtGui/QFont>
+#include <QtCore/QDebug>
+
+#include "SIMPLib/DataContainers/DataContainerArray.h"
+#include "SIMPLib/DataContainers/DataContainerArray.h"
```
mmarineBlueQuartz: Duplicate DataContainerArray includes

mmarineBlueQuartz (changes requested):
- FilterGroup00 needs to be examined to make sure a text substitution didn't go off the rails and rename the class unexpectedly.
- ArrayCalculator operator classes require updated formatting for their calculate methods.  Splitting the parameters between multiple lines but reducing the method brackets to one line alongside half the parameters makes it less clear than it should be if that is the method definition or an oddly placed lambda.

Odd but not important
- SIMPL_TYPE_MACRO(T) swapped out with duplicate method descriptions.
- Duplicated include paths in the same file.

Source/Experimental/FilterCallback/FilterParameter.h
```diff
@@ -258,6 +334,14 @@ class DataArrayPathParameter : public IFilterParameter
     DataArrayPathParameter(const std::string& humanLabel, const std::string& propertyName, const DataArrayPath &defaultValue, Category category, const SetterCallbackType& setterCallback, const GetterCallbackType& getterCallback, int groupIndex);
     
   private:
+    std::string m_HumanLabel = {};
+    std::string m_PropertyName = {};
+    Category m_Category = {};
+    int m_GroupIndex = {};
+    int m_DefaultValue = {};
+    double m_DefaultValue = {};
+    DataArrayPath m_DefaultValue = {};
```
mmarineBlueQuartz: I know this is in experimental, but I don't think you can have multiple variables of the same name and different types.

Source/SIMPLib/Common/CreatedArrayHelpIndexEntry.cpp
```diff
@@ -47,3 +47,136 @@ CreatedArrayHelpIndexEntry::CreatedArrayHelpIndexEntry()
...
+
+// -----------------------------------------------------------------------------
+CreatedArrayHelpIndexEntry::Pointer CreatedArrayHelpIndexEntry::New()
+{
+  Pointer sharedPtr(new(CreatedArrayHelpIndexEntry));
+  return sharedPtr;
+}
+
+// -----------------------------------------------------------------------------
+const QString CreatedArrayHelpIndexEntry::getNameOfClass() const
```
mmarineBlueQuartz: Unnecessary const in the beginning

Source/SIMPLib/Common/Constants.h
```diff
@@ -53,30 +65,29 @@
...
-//      TextFormat,
-//      XmlFormat,
-    UnknownFormat
-  };
+using Rgb = uint32_t;
+const Rgb RGB_MASK = 0x00ffffff; // masks RGB values
+const QString PathSep("|");
+static const uint8_t Unchecked = 0;
+static const uint8_t PartiallyChecked = 1;
+static const uint8_t Checked = 2;
```
mmarineBlueQuartz: `static const` global

Source/SIMPLib/Common/Constants.h
```diff
@@ -53,30 +65,29 @@
...
-    static const unsigned int LowtoHigh = 0;
-    static const unsigned int HightoLow = 1;
-    static const unsigned int UnknownRefFrameZDirection = 2;
-  }
+/** @brief Constants defined for the Stacking order of images into a 3D Volume */
+namespace RefFrameZDir
+{
+static const unsigned int LowtoHigh = 0;
+static const unsigned int HightoLow = 1;
+static const unsigned int UnknownRefFrameZDirection = 2;
```
mmarineBlueQuartz: `static const` globals

Source/SIMPLib/Common/CreatedArrayHelpIndexEntry.cpp
imikejackson: Those should be taken care of

Source/SIMPLib/Common/Constants.h
imikejackson: As we discussed, we are going to leave these for the time being until we can get on C++17 and use inline keyword

PR #370, opened 2019-07-09
Filling in IntVec2 widget values from the filter when preflight ends.

joeykleingers:
This allows filters to set recommended IntVec2 values during preflight.

Signed-off-by: Joey Kleingers <joey.kleingers@bluequartz.net>

### DIFF
--- a/Source/SVWidgetsLib/FilterParameterWidgets/IntVec2Widget.cpp
+++ b/Source/SVWidgetsLib/FilterParameterWidgets/IntVec2Widget.cpp
@@ -129,4 +129,10 @@ void IntVec2Widget::beforePreflight()
 // -----------------------------------------------------------------------------
 void IntVec2Widget::afterPreflight()
 {
+  if(getFilterParameter() != nullptr)
+  {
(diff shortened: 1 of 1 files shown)


imikejackson:
How do you not "fight" the preflight on this? If I fill in a value, and it preflights, wont my value get over written by the post-preflight values

joeykleingers:
When a filter preflights, it sets its IntVec2 property from the values in the user interface and calls dataCheck.  If the IntVec2 property stays unchanged in dataCheck, the widget will simply just re-set the same IntVec2 values into the user interface at the end of preflight.  However, if the filter decides to change the IntVec2 property during preflight to something that's recommended based on information parsed from a file or something else, those values would then be displayed in the user interface.  It would be up to each filter to decide whether or not it makes sense to set "recommended" values that the user interface will display at the end of preflight.  Obviously we don't want to overwrite what the user has put into the user interface, but there could be cases where we want to change the widget values to default to values that we found while parsing an input file.  For example, our montage import filters may want to set the Montage Start and Montage End IntVec2 properties whenever the cache changes so that the user interface defaults to the entire montage size that it reads during preflight.  However, those properties wouldn't get set if reading from an existing cache.

joeykleingers:
I don't know, maybe it's a bad idea.

PR #373, opened 2019-08-12
Feature/372 rename path

### LINKED ISSUE #372: RenameAttributeArrays does not play with with the renaming code
When using the rename attribute array filter in pipeline we get the issue where the preflight will mess up other filters array names and we get in that infinite loop of trying to update a filters inputs without having another filter updated.

mmarineBlueQuartz:
* Path renames are stopped when the original path is created again.
* Fixed cases where paths are infinitely renamed as one rename changed other paths created afterwards.
* Fixed a bug in SVPipelineView where getFilterPipeline() was not returning the expected result.
* Moved static rename functions from the DataArrayPath class to the RenameDataPath namespace.

fixes #372 
Requires PR BlueQuartzSoftware/DREAM3D#915


imikejackson (commented):
Do all the unit tests pass? Have you verified your fixes through manual testing in the UI?

mmarineBlueQuartz:
All unit tests pass.  All manual testing I could come up with in the UI seemed to work.  Rename filters saw improvements from the previous iteration in terms of renaming the output.  I ran through EBSD Reconstruction (08), and only special cases like FeatureIds and NumElements did not get renamed.

imikejackson:
Sounds good. Go ahead and merge.

PR #376, opened 2019-08-26
Enable Double Precision computations in some base classes.

imikejackson:


Source/SIMPLib/CoreFilters/ImportHDF5Dataset.cpp
```diff
@@ -250,7 +250,7 @@ void ImportHDF5Dataset::dataCheck()
       }
     }
     stream << "\n";
-    stream << tr("    Total HDF5 Dataset Element Count: %1\n").arg(locale.toString(hdf5TotalElements));
+    stream << tr("    Total HDF5 Dataset Element Count: \n") << hdf5TotalElements << "\n";
```
mmarineBlueQuartz: I think you meant to remove the first `/n`

Source/SIMPLib/Math/MatrixMath.h
```diff
@@ -265,29 +577,23 @@ class SIMPLib_EXPORT MatrixMath
...
-
-    /**
-     * @brief The dot product of 2 vectors a & b. Double Precision Variant
-     * @param a 1x3 Vector
-     * @param b 1x3 Vector
-     * @return
-     */
-    static float DotProduct3x1(double a[3], double b[3]);
+    template <typename T>
+    static float Magnitude4x1(T a[4])
```
mmarineBlueQuartz: This should probably return T, not float.

PR #378, opened 2019-10-13
Add filter that generates a tilt series of images from an ImageGeom

imikejackson:
Also a fix for a bug in ImageGeom when calculating the index of a voxel from a set of 3D Floating point coordinates.


Source/SIMPLib/CoreFilters/GenerateTiltSeries.cpp
```diff
@@ -0,0 +1,665 @@
...
+namespace Rotations
+{
+namespace Constants
+{
+#if DREAM3D_PASSIVE_ROTATION
+static const float epsijk = 1.0f;
+// static const double epsijkd = 1.0;
+#elif DREAM3D_ACTIVE_ROTATION
+static const float epsijk = -1.0f;
+static const double epsijkd = -1.0;
```
mmarineBlueQuartz: Avoid static global variables (not contained in class or function).  As of C++17, global constants should be purely const as it implies a single variable for compiler optimization, but what I've seen most commonly agreed upon is that static const global variables are inherently unsafe. The most common understanding I've seen is that each translation unit (.cpp) will have their own version of the variable.  Arguably, that's the same as const pre-17, but it looks like it's not, which can be dangerous.

I am not sure if static const globals post-17 break the assumptions for const globals, but there's a number of articles warning to not compare static const global variables with == operator due to differing memory addresses.  The pre-17 way to ensure you only have one version of a global is to use everyone's all-time favorite keyword, extern.

This is also a .cpp file, so none of that applies.  But we still should not be using static here.

Source/SIMPLib/CoreFilters/GenerateTiltSeries.cpp
```diff
@@ -0,0 +1,665 @@
...
+  }
+
+  // Compute a grid that forms a plane parallel to the rotation axis and centered halfway back in the volume
+  std::pair<FloatArrayType::Pointer, ImageGeom::Pointer> gridPair = generateGrid();
+  FloatArrayType::Pointer gridCoords = gridPair.first;
+  ImageGeom::Pointer gridGeometry = gridPair.second;
+  DataContainerArray::Pointer dca = getDataContainerArray();
+
+#ifdef SIMPL_USE_PARALLEL_ALGORITHMS
+  tbb::task_scheduler_init init;
```
mmarineBlueQuartz, line 476: Can we use ParallelTaskAlgorithm here?  Branching code paths is a code smell, and thread count is already checked in its execute as long as the algorithm object is created before the for loop.    If SIMPL_USE_PARALLEL_ALGORITHMS is not supported, it is not performed in parallel.

mmarineBlueQuartz (changes requested):
Overall looks good.  There were two points of interest, but I am mostly interested in the multi-threading part due to branching code paths around `#ifdef`.

Source/SIMPLib/CoreFilters/GenerateTiltSeries.cpp
imikejackson: Fixed

Source/SIMPLib/CoreFilters/GenerateTiltSeries.cpp
imikejackson: I don't think we can use ParallelTaskAlgorithm due to how the loops are developed with the rotations. I will continue to look at the code to see if we can shoe horn the algorithm into the ParallelTaskAlgorithm class.

mmarineBlueQuartz: It should just be along the lines of 
```
ParallelTaskAlgorithm alg;
alg.execute(Detail::ResampleGrid(this, gridCoords, gridDC, rotationAxis, gridIndex);
```
due to the fact that both versions call Detail::ResampleGrid's operator() to execute. Then call alg.wait() at the end.  Just make sure that `alg` is defined before the loop so it can keep track of its own threads.

mmarineBlueQuartz: If you can't use ParallelTaskAlgorithm, how can multi-threading be supported?  It's literally operating the same way as what's written in the loop?  Surely, they'd both have the same issue?

PR #380, opened 2019-10-28
Feature/const

JDuffeyBQ:
Changed appropriate functions in the following to be const:
* DataArray classes
* Geometry classes
* DataContainer
* AttributeMatrix
* DataContainerArray
* DynamicTableData

### DIFF
--- a/Source/SIMPLib/Geometry/IGeometry.h
+++ b/Source/SIMPLib/Geometry/IGeometry.h
@@ -225,7 +225,7 @@ class SIMPLib_EXPORT IGeometry : public Observable
      * @brief getElementsContainingVert
      * @return
      */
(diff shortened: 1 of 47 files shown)


Source/SIMPLib/Geometry/IGeometry.h
```diff
@@ -327,38 +327,38 @@ class SIMPLib_EXPORT IGeometry : public Observable
      * @brief getName
      * @return
      */
-    virtual QString getName() final;
+    virtual QString getName() const final;
```
mmarineBlueQuartz, line 330: I know you are not the one who put it in, but can we just comment on how absurd `virtual final` is?  That seems to be happening a lot in this class.

mmarineBlueQuartz (approved):

PR #382, opened 2019-11-13
Feature/generate tilt series fix

JDuffeyBQ:
Fixed gcc compile errors
* sqrtf isn't in the std namespace
* static constexpr member variable considered "undefined"


imikejackson (commented):
Looks good. Merge it in.

PR #384, opened 2019-11-18
Add DataArraySelectionFilterParameter::CreateRequirement for vector of strings

JDuffeyBQ:
* Can now use a static creation function to allow multiple types in DataArraySelectionFilterParameter instead of having to specify it in a separate statement

### DIFF
--- a/Source/SIMPLib/FilterParameters/DataArraySelectionFilterParameter.cpp
+++ b/Source/SIMPLib/FilterParameters/DataArraySelectionFilterParameter.cpp
@@ -161,6 +161,39 @@ DataArraySelectionFilterParameter::RequirementType DataArraySelectionFilterParam
 // -----------------------------------------------------------------------------
 //
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Source/SIMPLib/FilterParameters/DataArraySelectionFilterParameter.cpp
```diff
@@ -158,6 +158,39 @@ DataArraySelectionFilterParameter::RequirementType DataArraySelectionFilterParam
...
+    }
+  }
+
+  if(SIMPL::Defaults::AnyComponentSize != allowedCompDim)
+  {
+    req.componentDimensions = std::vector<SizeTVectorType>(1, SizeTVectorType(1, allowedCompDim));
+  }
+  if(AttributeMatrix::Type::Any != attributeMatrixType)
+  {
+    QVector<AttributeMatrix::Type> amTypes(1, attributeMatrixType);
```
mmarineBlueQuartz: use `AttributeMatrix::Types` instead of `QVector<AttributeMatrix::Type>`.  That way, if we ever change the alias to another collection type, this will get updated as well.

mmarineBlueQuartz (approved):

PR #386, opened 2019-12-23
Update Filter Parameter Widgets to use std::function callbacks instead of QMetaProperty

imikejackson:


mmarineBlueQuartz (changes requested):
This looks really good, and I like the new approach.  However, you might want to take a look at the change to FilterWidget.cpp (and remove it).

mmarineBlueQuartz (approved):

PR #393, opened 2020-03-19
Marking a few methods const that should not be changing the object

imikejackson:


imikejackson:
At a minimum, all tests pass on macOS with ITK 5, Python bindings

JDuffeyBQ (approved):

PR #398, opened 2020-05-01
Bug fix to initialize tuple when components > 1

megnashah:
This fixes an issue where I can initialize a tuple with more than one component.


imikejackson:
As discussed in person, the method does what it is intended to do. The method that should be used in the particular use case that we discussed is the "setTuple()"

PR #399, opened 2020-05-06
Allow user to specify the init values for each component of the tuple

imikejackson:

### DIFF
--- a/Source/SIMPLib/CoreFilters/CreateDataArray.cpp
+++ b/Source/SIMPLib/CoreFilters/CreateDataArray.cpp
@@ -104,7 +104,6 @@ std::vector<T> createInitVector(AbstractFilter* filter, const QString& initializ
 {
   using ConvertOutputType = std::pair<T, bool>;
   ConvertOutputType conversionResult;
-  convert_from<T> convert;
   std::vector<T> values(numComponents);
 
   if(initializationValue.startsWith(CDA_LB) && initializationValue.endsWith(CDA_RB))
@@ -113,52 +112,30 @@ std::vector<T> createInitVector(AbstractFilter* filter, const QString& initializ
     input = input.replace('[', "");
     input = input.replace(']', "");
     QStringList tokens = input.split(CDA_SEP, QString::SplitBehavior::KeepEmptyParts);
-    if(tokens.size() == 1)
+
+    for(int32_t i = 0; i < numComponents; i++)
     {
-      for(int32_t i = 0; i < numComponents; i++)
+      int32_t idx = (tokens.size() == 1 ? 0 : i);
+      conversionResult = convert_from<T>::convert(tokens[idx]);
+      if(conversionResult.second)
       {
-        conversionResult = convert.convert(tokens[0]);
-        if(conversionResult.second)
-        {
-          values[i] = conversionResult.first;
-        }
-        else
-        {
-          if(filter)
-          {
-            QString msg = QString("Input String '%1' could not be converted to a numeric value. ").arg(tokens[0]);
-            filter->setErrorCondition(-5400, msg);
-            values.clear();
-            return values;
-          }
-        }
+        values[i] = conversionResult.first;
       }
-    }
-    else
-    {
-      for(int32_t i = 0; i < numComponents; i++)
+      else
       {
-        conversionResult = convert.convert(tokens[i]);
-        if(conversionResult.second)
+        if(filter)
         {
-          values[i] = conversionResult.first;
-        }
-        else
-        {
-          if(filter)
-          {
-            QString msg = QString("Input String '%1' could not be converted to a numeric value.").arg(tokens[i]);
-            filter->setErrorCondition(-5400, msg);
-            values.clear();
-            return values;
-          }
+          QString msg = QString("Input String '%1' could not be converted to a numeric value. ").arg(tokens[0]);
+          filter->setErrorCondition(-5400, msg);
+          values.clear();
+          return values;
         }
       }
     }
   }
   else
   {
-    conversionResult = convert.convert(initializationValue);
+    conversionResult = convert_from<T>::convert(initializationValue);
     if(conversionResult.second)
     {
       values.assign(numComponents, conversionResult.first);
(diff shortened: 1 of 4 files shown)


Source/SIMPLib/CoreFilters/CreateDataArray.cpp
```diff
@@ -56,89 +56,97 @@
...
+      {                                                                                                                                                                                                \
+        return 0;                                                                                                                                                                                      \
+      }                                                                                                                                                                                                \
+      return value;                                                                                                                                                                                    \
+    }                                                                                                                                                                                                  \
+  };
+
+CDA_CREATE_CONVERTOR(uint8_t, toUInt)
```
JDuffeyBQ, line 82: All the non class functions can be put into an anonymous namespace.

Source/SIMPLib/CoreFilters/CreateDataArray.cpp
```diff
@@ -56,89 +56,97 @@
...
+CDA_CREATE_CONVERTOR(uint64_t, toULongLong)
+CDA_CREATE_CONVERTOR(int64_t, toLongLong)
+CDA_CREATE_CONVERTOR(bool, toInt)
+
+CDA_CREATE_CONVERTOR(float, toDouble)
+CDA_CREATE_CONVERTOR(double, toDouble)
+
+#define CDA_SEP ";"
+#define CDA_LB "["
+#define CDA_RB "]"
```
JDuffeyBQ:
```suggestion
constexpr char CDA_SEP = ';';
constexpr char CDA_LB = '[';
constexpr char CDA_RB = ']';
```
Change to constexpr char variables.

Source/SIMPLib/CoreFilters/CreateDataArray.cpp
```diff
@@ -56,89 +56,97 @@
...
+      {                                                                                                                                                                                                \
+        return 0;                                                                                                                                                                                      \
+      }                                                                                                                                                                                                \
+      return value;                                                                                                                                                                                    \
+    }                                                                                                                                                                                                  \
+  };
```
JDuffeyBQ:
```
#define CDA_CREATE_CONVERTOR(type, func)                                                                                                                                                               \
  template <>                                                                                                                                                                                          \
  struct convert_from<type>                                                                                                                                                                            \
  {                                                                                                                                                                                                    \
    static type convert(const QString& input)                                                                                                                                                          \
    {                                                                                                                                                                                                  \
      bool ok = false;                                                                                                                                                                                 \
      type value = static_cast<type>(input.func(&ok));                                                                                                                                                    \
      if(!ok)                                                                                                                                                                                          \
      {                                                                                                                                                                                                \
        return 0;                                                                                                                                                                                      \
      }                                                                                                                                                                                                \
      return value;                                                                                                                                                                                    \
    }                                                                                                                                                                                                  \
  };

template <class T>
struct convert_from
{
};

CDA_CREATE_CONVERTOR(uint8_t, toUInt)
CDA_CREATE_CONVERTOR(int8_t, toInt)
CDA_CREATE_CONVERTOR(uint16_t, toUShort)
CDA_CREATE_CONVERTOR(int16_t, toShort)
CDA_CREATE_CONVERTOR(uint32_t, toUInt)
CDA_CREATE_CONVERTOR(int32_t, toInt)
CDA_CREATE_CONVERTOR(uint64_t, toULongLong)
CDA_CREATE_CONVERTOR(int64_t, toLongLong)
CDA_CREATE_CONVERTOR(bool, toInt)
```

Using template specializations in the macro removes the need for the conversion template parameter in createInitVector(). The convert function can also be made static.

Source/SIMPLib/CoreFilters/CreateDataArray.cpp
```diff
@@ -429,71 +439,139 @@ void CreateDataArray::execute()
...
+    initializeArrayWithBool(m_OutputArrayPtr.lock(), m_InitializationType, values);
   }
 }
 
 // -----------------------------------------------------------------------------
 //
 // -----------------------------------------------------------------------------
-template <typename T> void CreateDataArray::checkInitialization(QString dataArrayName)
+template <typename T>
+void CreateDataArray::checkInitialization(QString dataArrayName)
```
JDuffeyBQ, line 524: Putting in a list of components fails for float and double.

Source/SIMPLib/CoreFilters/CreateDataArray.cpp
```diff
@@ -429,71 +439,139 @@ void CreateDataArray::execute()
...
 {
   if(m_InitializationType == Manual)
   {
     bool ok;
```
JDuffeyBQ: Should be initialized.

Source/SIMPLib/CoreFilters/CreateDataArray.cpp
```diff
@@ -429,71 +439,139 @@ void CreateDataArray::execute()
...
+      QStringList tokens = input.split(CDA_SEP, QString::SplitBehavior::KeepEmptyParts);
+      if(m_NumberOfComponents != tokens.size() && tokens.size() != 1)
+      {
+        QString ss = QString("Number of initialization components does not match number of components.%1 vs %2").arg(tokens.size()).arg(m_NumberOfComponents);
+        setErrorCondition(-5459, ss);
+        return;
+      }
+      for(const auto& token : tokens)
+      {
+        std::ignore = token.toDouble(&ok);
```
JDuffeyBQ: We should check the conversion with the actual type. For example if you enter -1 for an unsigned type, it the Qt functions fail to convert it resulting in 0, but it is a silent error during dataCheck().

JDuffeyBQ (changes requested):

imikejackson:
All the requested changes have been made.

Source/SIMPLib/CoreFilters/CreateDataArray.cpp
```diff
@@ -110,27 +117,58 @@ std::vector<T> createInitVector(const QString& initializationValue, int32_t numC
     {
       for(int32_t i = 0; i < numComponents; i++)
       {
-        values[i] = convert.convert(tokens[0]);
+        conversionResult = convert.convert(tokens[0]);
```
JDuffeyBQ:
```suggestion
        conversionResult = convert_from<T>::convert(tokens[0]);
```
Call statically here and in the other spots. Can remove local variables afterwards.

JDuffeyBQ (changes requested):

PR #400, opened 2020-05-06
DataArray iterators

JDuffeyBQ:
* Fixed issue with iterator's operator+ which was implemented as +=
* Extended iterator API to be a random access iterator
* Added cbegin()/cend()
* Added reverse iterators
* Changed uses of std::memcpy to std::copy
  * Changed parts of DataArray implementation to use iterators instead
* Changed templated begin()/end() to tupleBegin()/tupleEnd()

### DIFF
--- a/Source/SIMPLib/DataArrays/DataArray.cpp
+++ b/Source/SIMPLib/DataArrays/DataArray.cpp
@@ -367,8 +367,10 @@ typename DataArray<T>::Pointer DataArray<T>::WrapPointer(T* data, size_t numTupl
   // Allocate on the heap
   auto d = std::make_shared<DataArray<T>>(numTuples, name, compDims, static_cast<T>(0), false);
 
-  d->m_Array = data;        // Now set the internal array to the raw pointer
-  d->m_OwnsData = ownsData; // Set who owns the data, i.e., who is going to "free" the memory
+  // Now set the internal array to the raw pointer
+  d->m_Array = data;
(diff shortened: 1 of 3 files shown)


Source/SIMPLib/DataArrays/DataArray.cpp
```diff
@@ -208,9 +208,6 @@ DataArray<T>::DataArray(size_t numTuples, const QString& name, const comp_dims_t
     m_Size = m_NumTuples * m_NumComponents;
     m_MaxId = (m_Size > 0) ? m_Size - 1 : m_Size;
   }
-#if 0
-      MUD_FLAP_0 = MUD_FLAP_1 = MUD_FLAP_2 = MUD_FLAP_3 = MUD_FLAP_4 = MUD_FLAP_5 = 0xABABABABABABABABul;
```
imikejackson: This is handy to have in the code in case we are trying to figure out if somebody walked off the end of the array

Source/SIMPLib/DataArrays/DataArray.cpp
```diff
@@ -1447,17 +1457,8 @@ void DataArray<T>::deallocate()
       }
     }
   }
-#if 0
-      if (MUD_FLAP_0 != 0xABABABABABABABABul
```
imikejackson, line 1451: Let's think twice about getting rid of this little nugget of code. 99.9% of the time we don't need it, but when we do it is nice to just change the #if to 1 and recompile.

PR #401, opened 2020-05-13
Only copy public filters to the documentation directory

imikejackson:
Update codes to use this newer layout of "[PLUGIN]Filters" to
"[PLUGIN]_Filters" which allows mkdocs to insert a space character in the
TOC that is generated making the TOC more readable.

### DIFF
--- a/Source/SVWidgetsLib/QtSupport/QtSHelpUrlGenerator.cpp
+++ b/Source/SVWidgetsLib/QtSupport/QtSHelpUrlGenerator.cpp
@@ -35,9 +35,14 @@
 
 #include "QtSHelpUrlGenerator.h"
 
+#include <iostream>
+
+#include <QtCore/QTextStream>
 #include <QtCore/QCoreApplication>
 #include <QtCore/QDir>
+
 #include <QtGui/QDesktopServices>
+
 #include <QtWidgets/QApplication>
 #include <QtWidgets/QMessageBox>
 #include <QtWidgets/QWidget>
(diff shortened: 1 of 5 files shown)


Source/SVWidgetsLib/QtSupport/QtSHelpUrlGenerator.cpp
```diff
@@ -90,25 +95,18 @@ QUrl QtSHelpUrlGenerator::GenerateHTMLUrl(QString htmlName)
 
 #ifdef SIMPL_USE_MKDOCS
   {
-    FilterManager* fm = FilterManager::Instance();
-
-    IFilterFactory::Pointer factory = fm->getFactoryFromClassName(htmlName);
```
JDuffeyBQ: Without this, it fails to compile because factory is not defined.

JDuffeyBQ (changes requested):

Source/SVWidgetsLib/QtSupport/QtSHelpUrlGenerator.cpp
imikejackson: Interesting. It seemed to compile on my macOS machine but I also didn't do a completely clean compile.

JDuffeyBQ: Is SIMPL_USE_MKDOCS defined? It's in that ifdef block.

PR #402, opened 2020-05-19
Only copy Public filter docs into the build folder

imikejackson:
All the other repositories also have this same branch.


JDuffeyBQ (commented):
We'll need to add the superclass macro to ITKImageProcessing as well. The rest of the Python tests pass on my machine.

imikejackson:
I'll run another full clean build + testing on that and report back.

imikejackson:
If we go with a pure virtual for getNameOfClass() does that mess up the Pybind11 bindings?

JDuffeyBQ:
I don't think so since execute() is also pure virtual, and that hasn't caused any issues.

JDuffeyBQ:
I've compiled with getNameOfClass() as pure virtual and all the tests passed.

PR #404, opened 2020-05-28
Observer and JsonFilterParameterReader Python Bindings

imikejackson:


Source/SIMPLib/Common/Observer.cpp
```diff
@@ -42,6 +42,9 @@
 // -----------------------------------------------------------------------------
 Observer::Observer() = default;
 
+ Observer::Observer(QObject* parent)
+ : QObject(parent){}
+
```
JDuffeyBQ: I don't think we need this change anymore since we went with the new API in FilterPipeline.

Source/SIMPLib/Common/Observer.h
```diff
@@ -54,8 +54,18 @@ class SIMPLib_EXPORT Observer : public QObject, public IObserver
...
+  PYB11_SHARED_POINTERS(Observer)
+  PYB11_CREATION()
+  PYB11_END_BINDINGS()
+  // End Python bindings declarations
+
 public:
   Observer();
+
+  Observer(QObject* parent);
+
```
JDuffeyBQ: I don't think we need this change anymore since we went with the new API in FilterPipeline.

Source/SIMPLib/FilterParameters/JsonFilterParametersReader.h
```diff
@@ -51,6 +51,17 @@
  */
 class SIMPLib_EXPORT JsonFilterParametersReader : public AbstractFilterParametersReader
 {
+
+  // Start Python bindings declarations
+  PYB11_BEGIN_BINDINGS(JsonFilterParametersReader)
+  PYB11_SHARED_POINTERS(JsonFilterParametersReader)
+  PYB11_STATIC_NEW_MACRO(JsonFilterParametersReader)
+
+  PYB11_METHOD(FilterPipeline::Pointer readPipelineFromFile ARGS QString IObserver)
```
JDuffeyBQ:
```suggestion
  PYB11_METHOD(FilterPipeline::Pointer readPipelineFromFile)
```

Source/SIMPLib/Filtering/FilterPipeline.h
```diff
@@ -86,6 +86,8 @@ class SIMPLib_EXPORT FilterPipeline : public Observable
   PYB11_METHOD(bool empty)
   PYB11_METHOD(void clearErrorCode)
   PYB11_METHOD(void clearWarningCode)
+  PYB11_METHOD(void addObserver ARGS Observer)
+  PYB11_METHOD(void removeObserver ARGS Observer)
```
JDuffeyBQ, line 90:
```suggestion
  PYB11_METHOD(void addObserver)
  PYB11_METHOD(void removeObserver)
```

JDuffeyBQ (commented):
I tested on both my Windows and Linux machines, and everything compiled. And I ran a test Python pipeline with an observer which worked.

Also should the clang-format changes be in this PR?

PR #408, opened 2020-06-24
Ebsd API Updates (EbsdLib removed dependency on Qt5)

imikejackson:

### DIFF
--- a/Source/SIMPLib/DataArrays/StructArray.hpp
+++ b/Source/SIMPLib/DataArrays/StructArray.hpp
@@ -98,18 +98,15 @@ public:
     {
       return NullPointer();
     }
-    StructArray<T>* d = new StructArray<T>(numElements, true);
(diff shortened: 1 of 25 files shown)


Source/SIMPLib/DataArrays/StructArray.hpp
```diff
@@ -113,6 +113,33 @@ class StructArray : public IDataArray
...
+    {
+      if(d->Allocate() < 0)
+      {
+        // Could not allocate enough memory, reset the pointer to null and return
+        delete d;
+        return StructArray<T>::NullPointer();
+      }
+    }
+    d->setName(QString::fromStdString(name));
+    Pointer ptr(d);
```
mmarineBlueQuartz: If you store 'd' as a shared pointer initially, you don't have to manually delete it when returning NullPointer()

mmarineBlueQuartz (approved):

JDuffeyBQ (approved):

PR #412, opened 2020-08-07
Topic/rect grid geom get index

joeykleingers:
* Added getIndex methods to IGeometryGrid and subclasses.
* Added RectGridGeom unit test, initially testing new getIndex method.

### DIFF
--- a/Source/SIMPLib/Geometry/IGeometryGrid.h
+++ b/Source/SIMPLib/Geometry/IGeometryGrid.h
@@ -97,9 +97,7 @@ public:
   virtual void getCoords(size_t x, size_t y, size_t z, double coords[3]) const = 0;
   virtual void getCoords(size_t idx, double coords[3]) const = 0;
 
-  virtual std::optional<size_t> getIndex(float coords[3]) const = 0;
   virtual std::optional<size_t> getIndex(float xCoord, float yCoord, float zCoord) const = 0;
-  virtual std::optional<size_t> getIndex(double coords[3]) const = 0;
   virtual std::optional<size_t> getIndex(double xCoord, double yCoord, double zCoord) const = 0;
 
 public:
--- a/Source/SIMPLib/Geometry/IGeometryGrid.h
+++ b/Source/SIMPLib/Geometry/IGeometryGrid.h
@@ -97,6 +97,11 @@ public:
   virtual void getCoords(size_t x, size_t y, size_t z, double coords[3]) const = 0;
   virtual void getCoords(size_t idx, double coords[3]) const = 0;
 
+  virtual std::optional<size_t> getIndex(float coords[3]) const = 0;
+  virtual std::optional<size_t> getIndex(float xCoord, float yCoord, float zCoord) const = 0;
+  virtual std::optional<size_t> getIndex(double coords[3]) const = 0;
+  virtual std::optional<size_t> getIndex(double xCoord, double yCoord, double zCoord) const = 0;
+
 public:
   IGeometryGrid(const IGeometryGrid&) = delete;            // Copy Constructor Not Implemented
   IGeometryGrid(IGeometryGrid&&) = delete;                 // Move Constructor Not Implemented
--- a/Source/SIMPLib/Geometry/ImageGeom.cpp
+++ b/Source/SIMPLib/Geometry/ImageGeom.cpp
@@ -678,14 +678,6 @@ void ImageGeom::getCoords(size_t idx, double coords[3]) const
 // -----------------------------------------------------------------------------
 //
 // -----------------------------------------------------------------------------
-std::optional<size_t> ImageGeom::getIndex(float coords[3]) const
-{
-  return getIndex(coords[0], coords[1], coords[2]);
-}
-
-// -----------------------------------------------------------------------------
-//
-// -----------------------------------------------------------------------------
 std::optional<size_t> ImageGeom::getIndex(float xCoord, float yCoord, float zCoord) const
 {
   size_t x = (xCoord - (0.5f * m_Spacing[0]) - m_Origin[0]) / m_Spacing[0];
@@ -693,14 +685,6 @@ std::optional<size_t> ImageGeom::getIndex(float xCoord, float yCoord, float zCoo
   size_t z = (zCoord - (0.5f * m_Spacing[2]) - m_Origin[2]) / m_Spacing[2];
 
   return (m_Dimensions[1] * m_Dimensions[0] * z) + (m_Dimensions[0] * y) + x;
-}
-
-// -----------------------------------------------------------------------------
-//
-// -----------------------------------------------------------------------------
-std::optional<size_t> ImageGeom::getIndex(double coords[3]) const
-{
-  return getIndex(coords[0], coords[1], coords[2]);
 }
 
 // -----------------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


Source/SIMPLib/Geometry/ImageGeom.cpp
```diff
@@ -675,6 +675,46 @@ void ImageGeom::getCoords(size_t idx, double coords[3]) const
...
+}
+
+// -----------------------------------------------------------------------------
+//
+// -----------------------------------------------------------------------------
+std::optional<size_t> ImageGeom::getIndex(float xCoord, float yCoord, float zCoord) const
+{
+  size_t x = (xCoord - (0.5 * m_Spacing[0]) - m_Origin[0]) / m_Spacing[0];
+  size_t y = (yCoord - (0.5 * m_Spacing[1]) - m_Origin[1]) / m_Spacing[1];
+  size_t z = (zCoord - (0.5 * m_Spacing[2]) - m_Origin[2]) / m_Spacing[2];
```
mmarineBlueQuartz: Despite passing in floats, this method is performing double-precision when calculating x, y, and z. I normally would not say anything about such matters, but this method is a completely identical to `getIndex(double, double, double)`. Either stay consistent to the parameter precision or consolidate calculations into a single method. Copy-pasting should be avoided.

```suggestion
  size_t x = (xCoord - (0.5f * m_Spacing[0]) - m_Origin[0]) / m_Spacing[0];
  size_t y = (yCoord - (0.5f * m_Spacing[1]) - m_Origin[1]) / m_Spacing[1];
  size_t z = (zCoord - (0.5f * m_Spacing[2]) - m_Origin[2]) / m_Spacing[2];
```

Source/SIMPLib/Geometry/ImageGeom.h
```diff
@@ -321,6 +321,11 @@ class SIMPLib_EXPORT ImageGeom : public IGeometryGrid
   void getCoords(size_t x, size_t y, size_t z, double coords[3]) const override;
   void getCoords(size_t idx, double coords[3]) const override;
 
+  virtual std::optional<size_t> getIndex(float coords[3]) const override;
+  virtual std::optional<size_t> getIndex(float xCoord, float yCoord, float zCoord) const override;
+  virtual std::optional<size_t> getIndex(double coords[3]) const override;
+  virtual std::optional<size_t> getIndex(double xCoord, double yCoord, double zCoord) const override;
```
mmarineBlueQuartz: Methods don't need to be simultaneously virtual and override.

```suggestion
  std::optional<size_t> getIndex(float coords[3]) const override;
  std::optional<size_t> getIndex(float xCoord, float yCoord, float zCoord) const override;
  std::optional<size_t> getIndex(double coords[3]) const override;
  std::optional<size_t> getIndex(double xCoord, double yCoord, double zCoord) const override;
```

Source/SIMPLib/Geometry/RectGridGeom.h
```diff
@@ -261,6 +261,11 @@ class SIMPLib_EXPORT RectGridGeom : public IGeometryGrid
   void getCoords(size_t x, size_t y, size_t z, double coords[3]) const override;
   void getCoords(size_t idx, double coords[3]) const override;
 
+  virtual std::optional<size_t> getIndex(float coords[3]) const override;
+  virtual std::optional<size_t> getIndex(float xCoord, float yCoord, float zCoord) const override;
+  virtual std::optional<size_t> getIndex(double coords[3]) const override;
+  virtual std::optional<size_t> getIndex(double xCoord, double yCoord, double zCoord) const override;
```
mmarineBlueQuartz:
```suggestion
  std::optional<size_t> getIndex(float coords[3]) const override;
  std::optional<size_t> getIndex(float xCoord, float yCoord, float zCoord) const override;
  std::optional<size_t> getIndex(double coords[3]) const override;
  std::optional<size_t> getIndex(double xCoord, double yCoord, double zCoord) const override;
```

mmarineBlueQuartz (changes requested):
Don't mark methods as simultaneously virtual and override.
Avoid copy-pasting entire methods.

mmarineBlueQuartz (approved):

Source/SIMPLib/Geometry/IGeometryGrid.h
```diff
@@ -97,6 +97,11 @@ class SIMPLib_EXPORT IGeometryGrid : public IGeometry
   virtual void getCoords(size_t x, size_t y, size_t z, double coords[3]) const = 0;
   virtual void getCoords(size_t idx, double coords[3]) const = 0;
 
+  virtual std::optional<size_t> getIndex(float coords[3]) const = 0;
+  virtual std::optional<size_t> getIndex(float xCoord, float yCoord, float zCoord) const = 0;
+  virtual std::optional<size_t> getIndex(double coords[3]) const = 0;
+  virtual std::optional<size_t> getIndex(double xCoord, double yCoord, double zCoord) const = 0;
```
JDuffeyBQ:
```
virtual std::optional<size_t> getIndex(float coords[3]) const = 0;
virtual std::optional<size_t> getIndex(double coords[3]) const = 0
```

I know this is the same kind of signature as getCoords() has above, but do we need it? If so, it should probably be a std::array or a span (if we add that to the SDK) depending on the exact use case.

imikejackson: Or can we just use a float* with the documentation saying the pointer is a 3 element array? Sometimes we have millions/billions of coords during a runtime and I would rather not incur the penalty of dynamic casting? Thoughts.

JDuffeyBQ: This is where a span would be the best option especially since we're trying to move away from raw pointers.
`std::span<const float, 3>` is just a pointer and a length under the hood just with the size enforced at compile time.

Also `std::optional<size_t> getIndex(float coords[3]) const` just calls `std::optional<size_t> getIndex(float xCoord, float yCoord, float zCoord) const` anyways in the current implementation.

mmarineBlueQuartz: `std::optional<size_t> getIndex(float coords[3]) const` is treated as `std::optional<size_t> getIndex(float* coords) const` by the compiler but is more descriptive to the user about what is expected. Making the code less descriptive and more reliant on comments is going the wrong way.

I don't feel strongly about span, but it does seem to be the next step in improving upon how we pass data.

imikejackson: I just want to make sure that whatever we chose is the "right" way update the API. My guess is that if we decide to update the API these are non-trivial changes to the dependent codes. I could be wrong. there may not be a lot of filters that depend on these APIs.

imikejackson: I'm for having expressive APIs but not at the detriment to the developer understanding of the API itself.

joeykleingers: So which way are we going with this.  We need to make a decision so that we can merge it in.

JDuffeyBQ: We don't necessarily have to update the old API, and we can just add the new getIndex() methods. We could discuss at the meeting as well.

joeykleingers: We discussed at the meeting and decided to remove the array method declarations for now.

PR #423, opened 2021-08-11
Improve Python Example use_case_02.py with additional example codes

mhitzem:
- Generates a new csv file with random data (fake scalar data)
- Data from file is read and put into uint32 and float32 arrays and added to AttributeMatrix and DataContainer


Wrapping/Python/Examples/use_case_02.py
```diff
@@ -36,6 +45,35 @@ def import_csv_xyz_vertices_2(file_path: str):
     dca.addOrReplaceDataContainer(dc)
     # Write the Vertex Geometry out to a .dream3d file.
     err = simpl_helpers.WriteDREAM3DFile('vert_geom_2.dream3d', dca, True)
+    
+    # Create Cell AttributeMatrix
+    tupleDims = simpl.VectorSizeT([1, 37989])
+    am = simpl_helpers.CreateAttributeMatrix(tupleDims, 'CellAttributeMatrix', simpl.AttributeMatrix.Type.Cell)
+
```
imikejackson, line 58: Probably need to call your ```make_random_data()``` function to actually create the file.

Wrapping/Python/Examples/use_case_02.py
```diff
@@ -11,7 +11,16 @@
 import dream3d.simpl_helpers as simpl_helpers
 
 
-
+#Generates a csv file with floats 0-37988 and ints 1-37989
+def make_random_data():
```
imikejackson: Could we add an argument to the function `num_elements` so that callers of the data can set the number of values to be written to the csv file?
Also we might want to add an argument for the output file so that a user of the function can set the output file.

Wrapping/Python/Examples/use_case_02.py
```diff
@@ -36,6 +45,35 @@ def import_csv_xyz_vertices_2(file_path: str):
     dca.addOrReplaceDataContainer(dc)
     # Write the Vertex Geometry out to a .dream3d file.
     err = simpl_helpers.WriteDREAM3DFile('vert_geom_2.dream3d', dca, True)
+    
+    # Create Cell AttributeMatrix
+    tupleDims = simpl.VectorSizeT([1, 37989])
```
imikejackson: Where did the '37989' value come from? this value should match the number of vertices read from the input file.

Maybe something like
```
    numVerts = len(verts_np)
    # Create Cell AttributeMatrix
    tupleDims = simpl.VectorSizeT([1, numVerts])
```

Wrapping/Python/Examples/use_case_02.py
```diff
@@ -36,6 +45,35 @@ def import_csv_xyz_vertices_2(file_path: str):
...
+    am = simpl_helpers.CreateAttributeMatrix(tupleDims, 'CellAttributeMatrix', simpl.AttributeMatrix.Type.Cell)
+
+    # Create Data Arrays
+    #Loads in csv file and gets data for each type
+    uints = np.genfromtxt('new_data.csv', dtype=np.uint32, skip_header=1, usecols=1, delimiter=',')
+    f32s = np.genfromtxt('new_data.csv', dtype=np.float32, usecols=0, skip_header=1, delimiter=',')
+
+    #Creates arrays from data with shape of (# of vertices x 1)
+    uintarr = np.ndarray(shape=(37989, 1), dtype=np.uint32, buffer=uints)
+    floatarr = np.ndarray(shape=(37989, 1), dtype=np.float32, buffer=f32s)
```
imikejackson: use the `numVerts` variable from above instead of the 37989 hard coded value.

Wrapping/Python/Examples/use_case_02.py
```diff
@@ -36,6 +45,35 @@ def import_csv_xyz_vertices_2(file_path: str):
     dca.addOrReplaceDataContainer(dc)
     # Write the Vertex Geometry out to a .dream3d file.
     err = simpl_helpers.WriteDREAM3DFile('vert_geom_2.dream3d', dca, True)
```
imikejackson: This line needs to be moved to the bottom of the function so that everything that got added to the DataStructure hierarchy is written to the DREAM3D file.

Wrapping/Python/Examples/use_case_02.py
```diff
@@ -36,6 +45,35 @@ def import_csv_xyz_vertices_2(file_path: str):
...
+    
+    # Create Cell AttributeMatrix
+    tupleDims = simpl.VectorSizeT([1, 37989])
+    am = simpl_helpers.CreateAttributeMatrix(tupleDims, 'CellAttributeMatrix', simpl.AttributeMatrix.Type.Cell)
```
imikejackson: We told you the incorrect type of AttributeMatrix to create. It should be a "vertex"
``` simpl.AttributeMatrix.Type.Vertex```

**Review by imikejackson — COMMENTED** [review]
Also, could we also add an argument "--output_dir" to the main args that allows the user to set where all of the output files are created. Be sure to flow that through all needed parts of the code.

File: Wrapping/Python/Examples/use_case_02.py [hunk 7]
```diff
@@ -116,7 +123,8 @@ def import_csv_xyz_vertices_1(file_path: str):
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
     parser.add_argument('--input_file')
+    parser.add_argument('--output_file')
```
imikejackson: This should be 'output_dir' and any output file that is written (.csv and dream3d files), those names should be prefixed with the output_dir path that is passed into the program

Wrapping/Python/Examples/use_case_02.py
```diff
@@ -43,38 +44,44 @@ def import_csv_xyz_vertices_2(file_path: str):
...
-    f32arr = simpl.FloatArray(floatarr, 'Float32 Arr', True)
-    
     # Add Data Arrays to AttributeMatrix
     am.addOrReplaceAttributeArray(uint32arr)
     am.addOrReplaceAttributeArray(f32arr)
 
     # Add AttributeMatirx to DataContainer
     dc.addOrReplaceAttributeMatrix(am)
 
+    err = simpl_helpers.WriteDREAM3DFile('vert_geom_2.dream3d', dca, True)
```
imikejackson: The output file path should be a concatenation of the output_dir variable the actual name of the output file, in this instance 'vert_geom_2.dream3d'.

Please update all places in the code where output files are written.

Wrapping/Python/Examples/use_case_02.py
```diff
@@ -34,9 +44,44 @@ def import_csv_xyz_vertices_2(file_path: str):
...
+    #     Create 2 Data Arrays: UInt32 and Float32
+    #     Populate data ararys with some random data
+    #       or
+    #     for bonus points read the data from another text file
+    # Add Data Arrays to AttributeMatrix
+    am.addOrReplaceAttributeArray(uint32arr)
+    am.addOrReplaceAttributeArray(f32arr)
+
+    # Add AttributeMatirx to DataContainer
+    dc.addOrReplaceAttributeMatrix(am)
```
imikejackson, line 83: See if you can add the DREAM3D filter "apply_transformation_to_geometry" which is in the DREAM3DReview plugin. Take a look at [https://github.com/dream3d/DREAM3DReview/blob/develop/Test/Python/Apply_Transformation_To_Geometry.py](https://github.com/dream3d/DREAM3DReview/blob/develop/Test/Python/Apply_Transformation_To_Geometry.py) file for inspiration.

In that example we are using a transformation matrix but see if you can figure out how to use just a scaling factor in each direction. Try out a scaling factor of 10x in each direction just to make it obvious when the data is viewed.

Write out the dream3d file before the transformation and after then use [ParaView](https://www.paraview.org/download) to view the results and verify that you have successfully scaled the vertices.

PR #424, opened 2021-11-01
Documentation for creating a python filter

mhitzem:


imikejackson:
Overall looks good. Generally when presenting code snippets you would fence off the section with a 
\```(lang-python)

\``` 
section

PR #430, opened 2022-08-15
Feature/copy feature array to element array

VKUDRI:
Altered error conditions so instances with less Feature IDs than the total available are still valid

### DIFF
--- a/Source/SIMPLib/CoreFilters/CopyFeatureArrayToElementArray.cpp
+++ b/Source/SIMPLib/CoreFilters/CopyFeatureArrayToElementArray.cpp
@@ -220,7 +220,7 @@ void CopyFeatureArrayToElementArray::execute()
 
   if(largestFeature < (numFeatures - 1))
   {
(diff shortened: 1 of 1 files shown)


imikejackson (changes requested):
I think we would need to get a "Fill value" from the user for the unused indices in the feature level array. We could just assume zero but that has caused problems in the past.

VKUDRI:
> I think we would need to get a "Fill value" from the user for the unused indices in the feature level array. We could just assume zero but that has caused problems in the past.

Would there be unused indices? If I have feature ids  1-7 and then my feature array has instances of 1,3,4,5, and 6 all the indices in the created element array will have valid ids, it just will not use all the possible ids.

imikejackson:
Sounds good then.

imikejackson (approved):

VKUDRI:
Changed error message, converted to warning

imikejackson:
These fixes were incorporated into another PR. Closing.

